PR #57, opened 2026-03-02
Refactor Graph/GraphInfo to use an enum for the operator options

mtavenrath:
With a string base interface introducing snake_case / camelCase errors is extremely easy and hard to detect. Introduce an enum instead for the operator options for compile time type and attribute name checking.

TRTX tests are passing.


mtavenrath:
Direct RustNN WPT ONNX tests are passing. rustnnpt tests will follow next

Issue #56, opened 2026-02-27
Checkin: using Rustnn in a Rust WebNN implementation


gterzian:
Hello @tarekziade 

This is just a heads-up that I am using this libary to implement WebNN in (a fork of) Servo (it's a fork because at the same time it's a personal experiment with AI coding and that is banned upstream). 

You can see the latest code over at https://github.com/gterzian/servo/compare/master...gterzian:servo:webnn

I have only a few operands so far, but the the full graph building and dispatching pipeline is present, and executes on coreml. 

Besides being just FYI, I think there might be an opportunity to have some kind of shared backend between pywebnn and what I'm doing. Effectively the Servo code in `components/script/dom/webnn` is the equivalent of the part of pywebnn that integrates with Python, and the part that could be shared would be in `components/webnn`. 

I haven't ever integrated Rust in Python, but I assume using threads in Rust is a way to get around the GIL(which I assume still exists?), and so running the compilation and compute on a backend as in `components/webnn` could be a step-up from [running it on an asyncio event-loop](https://github.com/rustnn/pywebnn/blob/b4ffb5c9c68538914c6847d05e7e113c15202935/python/webnn/__init__.py#L162) (where I assume threads still do not give you parallelism). 

In a Web Engine you want at least the option of doing that in a separate process, which the Servo code supports, but for the Python use-case I assume the code is trusted and so threading would be enough? 

I'm still in the early prototyping stage, but eventually it might make sense to have some kind of project under the rustnn org that would provide the compute backend for a runtime (Python or Web). 

Could also be interesting to use the fork of Servo containing the WebNN implementation to run WPT tests. 

Please let me know what you think.

tarekziade:
Thanks for the fix @gterzian — I’ll comment on the PR.

I should also mention that @mtavenrath has been working on a heavy refactoring around structs, but hopefully it won’t interfere too much with your fix.

I also saw your post about your AI experiment, it was really nice to see you experimenting with RustNN!

> I have only a few operands so far, but the full graph building and dispatching pipeline is present, and executes on coreml.

Really cool!

> I haven't ever integrated Rust in Python, but I assume using threads in Rust is a way to get around the GIL (which I assume still exists?), and so running the compilation and compute on a backend as in components/webnn could be a step-up
> from running it on an asyncio event-loop (where I assume threads still do not give you parallelism).

[...]

PR #58, opened 2026-03-03
Fix casing for argMax and argMin in coreml converter

gterzian:
## Summary

Since the match is on the lower case transformation of the operand name, the upper case "argMin" and "argMax" are never matched on and the converted program is missing information, which makes subsequent compilation fails. 

## Validation

- [x] `make test`

I've run into this problem while working on https://github.com/rustnn/rustnn/issues/56
and this was the only way for me to get a subset of `/wpt/tests/webnn/conformance_tests/arg_min_max.https.any.js` to pass on CoreML(I'm assuming the results at https://rustnn.github.io/rustnnpt/ are only testing ONNX?)

### DIFF
--- a/src/converters/coreml_mlprogram.rs
+++ b/src/converters/coreml_mlprogram.rs
@@ -1922,7 +1922,7 @@ impl CoremlMlProgramConverter {
                 }
             }
 
-            "argMax" | "argMin" => {
+            "argmax" | "argmin" => {
                 // reduce_argmax/reduce_argmin: x, axis, keep_dims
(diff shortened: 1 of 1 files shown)


tarekziade:
@gterzian I extended `rustnnpt` so it can run against CoreML or any backend. see the `main` README 

CoreML crashes on many tests, (I guess the non compatible quantized levels) but that's now captured so we can run all tests

```
npm run test:wpt:run -- --backend coreml --variants npu --runner-features backend-onnx,backend-coreml
```

PR #77, opened 2026-03-18
make model compilation function public

gterzian:
## Summary

This is the only difference with my fork; 

In the Servo implementation, I have separated the building from the dispatching, and as part of building I:

- convert the graph to a coreml program
- compile the model
-  see [here](https://github.com/rustnn/servo-ml/blob/9d1aa4db00c00bf2c6c065345f2f6f59f270a3b6/components/webnn/src/lib.rs#L787)

I then store the graph, the converted program, and the path to the model in a kind of memory only cache(which could be persisted to disk), and later as part of dispatch I [call into `run_coreml_with_inputs_cached`](https://github.com/rustnn/servo-ml/blob/9d1aa4db00c00bf2c6c065345f2f6f59f270a3b6/components/webnn/src/lib.rs#L959) with those values from the cache. 

In comparison, in pywebnn, for coreml, [everything is done at dispatch time](https://github.com/rustnn/pywebnn/blob/8b68c8cab61349758509e00d02f5fb26fe71a14c/src/python/context.rs#L1301).

At the end of the day, I think this only matters in terms of potential UI. 

If the building of a graph is very fast, because it doesn't  run the conversion or compiles the model, and then the dispatch is slower (the first time) because it needs to do all these steps, then I think that can be unexpected to a user. 

On the other hand, showing some kind of pretty slow progress bar while the build is ongoing might be more acceptable. 

The spec also does the conversion as part of the build method, see the stuff starting at step 18 of https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-build

Please let me know what you think.

## Validation

- [x] `make test`

## Documentation


tarekziade:
> > Can I just enable auto-merge?
> 
> Yes, give it a try. If it ever causes trouble we'll disable the feature.

+1 , good idea - I will also tweak how I review (sometimes I approve with nits, so I will just nit first)

PR #88, opened 2026-04-09
refactor: add ML.*Limits types

theHamsta:
Prior art in
https://github.com/rustnn/pywebnn/blob/91ac33a28de47ee6ea9658af852cc7f898a88a4f/src/python/context.rs#L722

Those were generated using AI from WebIDL and wasm_sys files

## Summary

Attempt to bring more types from WebNN api into RustNN. Since the concept does not exist yet I started with Limits.

The limits can be obtained as a single dictionary attached to a MLContext. Each MLContext (for us a backend+device combo) can specify their limits. This just adds the types.

Since AI assisted, there might be errors in types. Will review/cross check with other AI

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/limits/mod.rs
+++ b/src/limits/mod.rs
@@ -1,0 +1,313 @@
+//! WebNN [`MLOpSupportLimits`](https://www.w3.org/TR/webnn/#dictdef-mlopsupportlimits) and related
+//! limit dictionaries, for JSON interchange and alignment with `web-sys` / wasm-bindgen bindings.
+
+mod types;
+
+pub use types::{
+    MLBatchNormalizationSupportLimits, MLBinarySupportLimits, MLConcatSupportLimits,
+    MLConv2dSupportLimits, MLGatherSupportLimits, MLGemmSupportLimits, MLGruCellSupportLimits,
+    MLGruSupportLimits, MLLogicalNotSupportLimits, MLLstmCellSupportLimits, MLLstmSupportLimits,
+    MLNormalizationSupportLimits, MLPreluSupportLimits, MLQuantizeDequantizeLinearSupportLimits,
+    MLScatterSupportLimits, MLSingleInputSupportLimits, MLSplitSupportLimits, MLWhereSupportLimits,
+};
+
+use serde::{Deserialize, Serialize};
+
+use crate::graph::DataType;
+
+/// `MLInputOperandLayout` — preferred layout for layout-dependent operators (e.g. conv2d).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum MLInputOperandLayout {
+    Nchw,
+    #[default]
+    Nhwc,
+}
+
+/// `MLRankRange` — inclusive min/max tensor rank supported.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MLRankRange {
+    pub min: u32,
+    pub max: u32,
+}
+
+/// `MLTensorLimits` — allowed operand data types and rank range.
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MLTensorLimits {
+    pub data_types: Vec<DataType>, // use &[DataType] ?
+    pub rank_range: MLRankRange,
+}
+
+/// `MLOpSupportLimits` — merged dictionary from the specification (base + all partials).
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct MLOpSupportLimits {
+    pub preferred_input_layout: MLInputOperandLayout,
+    pub max_tensor_byte_length: u64,
+    pub input: MLTensorLimits,
+    pub constant: MLTensorLimits,
+    pub output: MLTensorLimits,
+
+    #[serde(default)]
+    pub arg_min: Option<MLSingleInputSupportLimits>,
+    #[serde(default)]
+    pub arg_max: Option<MLSingleInputSupportLimits>,
+    #[serde(default)]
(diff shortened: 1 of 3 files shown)


mtavenrath:
Split into smaller structs to be more flexible.

matthargett:
this seems more important now that the May WebNN CRD makes opSupportLimits() the mechanism for choosing a model architecture for a target implementation ([§8.3.7](https://www.w3.org/TR/2026/CRD-webnn-20260521/#api-mlcontext-opsupportlimits)). One thing I’d want to avoid is two schema sources now that #206  vendors the generated WebNN IDL bindings. could these core structs be generated from or round-tripped against that same IDL? The useful next slice on watchOS seems to be wiring this into MLContext::op_support_limits() with one backend’s real rank/type limits—CoreML would exercise the design immediately. 

is there anything I can do to help get this rebased, validated, and moving forward? or should I kick off a new PR?

theHamsta:
> this seems more important now that the May WebNN CRD makes opSupportLimits() the mechanism for choosing a model architecture for a target implementation ([§8.3.7](https://www.w3.org/TR/2026/CRD-webnn-20260521/#api-mlcontext-opsupportlimits)). One thing I’d want to avoid is two schema sources now that #206 vendors the generated WebNN IDL bindings. could these core structs be generated from or round-tripped against that same IDL? The useful next slice on watchOS seems to be wiring this into MLContext::op_support_limits() with one backend’s real rank/type limits—CoreML would exercise the design immediately.
> 
> is there anything I can do to help get this rebased, validated, and moving forward? or should I kick off a new PR?

@matthargett sorry, for seeing this delayed. I was on PTO. Feel free to create a new PR.

If you think that generating the public API from IDL would be beneficial you could look into https://github.com/wasm-bindgen/wasm-bindgen/tree/main/crates/web-sys https://github.com/wasm-bindgen/wasm-bindgen/tree/main/crates/webidl . The parser there can parse the IDL files for WebNN and it generated me the wasm bindings for browser WebNN. It could be changed to produce a rust api instead of just the JS bindings.

theHamsta:
@matthargett since this is just adding the types... Should we just merge and you add the required API functions on top of it?

src/limits/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct MLOpSupportLimits {
+    pub preferred_input_layout: MLInputOperandLayout,
+    pub max_tensor_byte_length: u64,
+    pub input: MLTensorLimits,
+    pub constant: MLTensorLimits,
+    pub output: MLTensorLimits,
+
+    #[serde(default)]
+    pub arg_min: Option<MLSingleInputSupportLimits>,
```
theHamsta, line 54: to Option or not to Option? I got those struct via the IDL builder and that one generated Option. A backend should return preferably limits for all options. So, remove Option here and set Defaults with `min: 0, max: u32::MAX`?

mtavenrath: to Option if this is what's defined by the IDL. For RustNN internal usage we can provide a function get*Limits(Option<*Limits>) ->*Limits which returns reasonable limits.

theHamsta: well the IDL has

```
dictionary MLOpSupportLimits {
  MLInputOperandLayout preferredInputLayout;
  [EnforceRange] unsigned long long maxTensorByteLength;
  MLTensorLimits input;
  MLTensorLimits constant;
  MLTensorLimits output;
};

dictionary MLRankRange {
  unsigned long min;
  unsigned long max;
};

typedef sequence<MLOperandDataType> MLDataTypeList;

dictionary MLTensorLimits {
  MLDataTypeList dataTypes;
  MLRankRange rankRange;
};

dictionary MLBinarySupportLimits {
  MLTensorLimits a;
  MLTensorLimits b;
  MLTensorLimits output;
};
```
the wasm_bindgen generator generated gettter and setter with optionals, because Javascript. Also, because of future extensions of WebNN or old WebNN, your Javascript app effectively might get non-present entries in this object.

Since this object is only ever returned by RustNN, we could guarantee that the entries are always present, which also looks a bit more like the IDL. 

We should only return references to a static reference of this struct to not copy us to death by the size of the struct.

theHamsta: I now remember that I had the idea that we could accept `MLOpSupportLimits` also from the user when they would convert ONNX files or write WebNN files.

This way they could make sure that their model would fit into a certain limit they want to be supported on all major browsers and backends.

src/limits/mod.rs
```diff
@@ -0,0 +1,449 @@
...
+                MLOperandDataType::Uint64,
+                MLOperandDataType::Int8,
+                MLOperandDataType::Uint8,
+                MLOperandDataType::Int4,
+                MLOperandDataType::Uint4,
+            ],
+            rank_range: Default::default(),
+        }
+    }
+}
```
theHamsta: adding Default for those types in opinionated, but should make implementations for backends easier and allow it for users to provide their own limit from a serialization.

We can provide global limits here as Default and backends can restrict where they feel that they can't support a specific thing. Long-term all backends should get very close to the Default in the spirit to resist fingerprinting via limits. (Spec recommendation recommends to provide few generic limit profiles and apply to as many clients as possible)

src/limits/mod.rs
```diff
@@ -0,0 +1,449 @@
...
+                MLOperandDataType::Float16,
+                MLOperandDataType::Int32,
+                MLOperandDataType::Uint32,
+                MLOperandDataType::Int64,
+                MLOperandDataType::Uint64,
+                MLOperandDataType::Int8,
+                MLOperandDataType::Uint8,
+                MLOperandDataType::Int4,
+                MLOperandDataType::Uint4,
+            ],
```
theHamsta: this could also be a `&static [MLOperandDataType]`. Then, one would not have to have tens of static allocations per `MLOpSupportLimits`. Again we really want `MLOpSupportLimits` for each backend only be generated once and then return an immutable reference to the user

mtavenrath: What if a backend has supported for multiple devices, each with a different set different limits?

theHamsta: > What if a backend has supported for multiple devices, each with a different set different limits?

Then, you should return the minimum support device limits. You can generate limits per context which would be per device

PR #90, opened 2026-04-10
refactor: add strong enums

theHamsta:
This commit just adds the enums and does not use them yet. We can do small refactors to use those enums one-by-one in smaller changes

## Summary

Add enums from the WebNN specs (currently we use strings like the JS API). Just add the enums, use them in follow-up PRs.

## Validation

- [ ] `make test`

## Documentation


mtavenrath (approved):
LGTM

PR #95, opened 2026-04-20
refactor: run multiple runs of `cargo clippy --fix`

theHamsta:
This fixes the clippy issues that can be fixed automatically.

## Summary

## Validation

- [ ] `make test`

## Documentation


src/converters/trtx.rs
```diff
@@ -10789,7 +10778,7 @@ impl TrtxConverter {
             });
         }
 
-        let sh = opts.strides.get(0).copied().unwrap_or(1);
+        let sh = opts.strides.first().copied().unwrap_or(1);
```
theHamsta, line 10781: not a fan of this lint when there is also `get(1)`. should I allow this clippy lint?

mtavenrath: I am undecided. I guess it's better to keep the default?

PR #96, opened 2026-04-20
tests: Try to generate separate tests for WPT jsons

theHamsta:
## Summary

I would like to either have different Rust tests for each JSON (will run in parallel) to have an easy overview and filter out using `cargo test` and/or use snap shot testing via https://insta.rs/docs/cli/ to track improvements or regressions 

Since a lot of tests are failing, would do snapshot testing first since less invasive (this change is AI generated)

## Validation

- [ ] `make test`

## Documentation


mtavenrath:
WPT json tests were introduces as first try to run WPT tests. In the long term they'll be deprecated in favor of rustnnpt or an even more complete node.js based test runner which also allows tests selections.

We want to be able to integrate those javascript runners into the rust testing infrastructure.

PR #99, opened 2026-04-21
refactor: use MLOperandDataType

theHamsta:
I grepped the generated API and it seems to be only used for MLArgMinMaxOptions and as an argument to cast. Of course, it is also used in the general high-level API (MLTensor) where we currently mostly use rustnn::DataType

## Summary

## Validation

- [ ] `make test`

## Documentation


src/converters/onnx.rs
```diff
@@ -10031,7 +10016,7 @@ mod tests {
     }
 
     #[test]
-    fn test_cast_operation_with_int4_target() {
+    fn test_cast_operation_to_int32() {
```
theHamsta, line 10019: would take away int4 for now

PR #100, opened 2026-04-21
Add draft for MLContext API

theHamsta:
## Summary

Attempt for a high-level API resembling WebNN, similar to what we already have in pywebnn. Need to decide whether 

Goal is to have context object that backends could implement to keep their resources alive and destroy them accordingly. A similar object will be needed for MLTensor, MLGraph. MLGraphBuild could be by passed for quick iteration by our serialization formats/converter infra.

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/mlcontext.rs
+++ b/src/mlcontext.rs
@@ -3,11 +3,12 @@
 #[cfg(any(feature = "trtx-runtime", feature = "trtx-runtime-mock"))]
 use crate::executors::trtx::TrtxGraph;
 use crate::{
+    GraphInfo,
     backend_selection::BackendDevice,
     error::{Error, Result},
     executors::trtx::TrtxContext,
 };
-use std::{collections::HashMap, fmt::Display, io::Read};
+use std::{collections::HashMap, fmt::Display};
 
 use crate::{
     backend_selection::{select_backend, select_backend_by_gpu},
@@ -45,7 +46,8 @@ pub(crate) trait MLBackendContext<'context>: std::fmt::Debug {
 
 pub(crate) trait MLBackendBuilder<'context>: std::fmt::Debug {
     /*async*/
-    fn build(&self) -> Result<MLGraph<'context>>;
+    fn build(&mut self, outputs: &HashMap<&str, MLOperand>) -> Result<MLGraph<'context>>;
+    fn load_graph(&mut self, webnn: &'context GraphInfo) -> Result<()>;
 }
 
 // can be made a Box<dyn better_any::Tid<'context> + 'context> for dynamic dispatch
@@ -201,6 +203,11 @@ pub struct MLTensorDescriptor {
     writable: bool,
 }
 
+#[derive(Debug, Eq, PartialEq, Default, Clone)]
+pub struct MLOperand {
+    pub(crate) id: usize,
+}
+
 impl std::ops::Deref for MLTensorDescriptor {
     type Target = MLOperandDescriptor;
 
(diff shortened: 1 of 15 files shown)


src/mlcontext.rs
```diff
@@ -0,0 +1,153 @@
...
+// Backend traits
+
+// could make public later if interface stabilized
+pub(crate) trait MLBackendContext: std::fmt::Debug {
+    fn accelerated(&self) -> bool;
+}
+
+pub(crate) trait MLBackendTensor: std::fmt::Debug {}
+
+// type for MLContext
```
theHamsta: TODO: add creation methods for context. For WebNN they are on the `ML` objects. We could make them globals or static members on MLContext

src/mlcontext.rs
```diff
@@ -0,0 +1,153 @@
...
+pub struct MLOpSupportLimits {}
+
+// https://www.w3.org/TR/webnn/#dictdef-mloperanddescriptor
+#[derive(Debug, Eq, PartialEq)]
+pub struct MLOperandDescriptor {
+    data_type: MLOperandDataType,
+    shape: Vec<u64>,
+}
+
+impl MLOperandDescriptor {
```
theHamsta, line 159: or just POD, like JS

src/mlcontext.rs
```diff
@@ -0,0 +1,153 @@
...
+    // omit destroy()? We're not JS, objects can be destroyed via drop, we could do destroy stuff in Drop
+    pub fn destroy(self) {
+        todo!()
+    }
+
+    pub fn dispatch(
+        &mut self,
+        graph: &MLGraph,
+        inputs: &HashMap<&str, MLTensor>,
+        outputs: &HashMap<&str, MLTensor>,
```
theHamsta: needs probably lifetime annotations.

theHamsta:
TODO: we need to return a `Result` for all fallible methods. We currently have `GraphError` as only Error type that is used for everything including inference.

src/mlcontext.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::operator_enums::MLOperandDataType;
+
+// Backend traits
+
+// could make public later if interface stabilized
+pub(crate) trait MLBackendContext: std::fmt::Debug {
+    fn accelerated(&self) -> bool;
+}
+
+pub(crate) trait MLBackendTensor: std::fmt::Debug {}
```
theHamsta: or just have an `id` in MLTensor, since everything with tensors goes through `MLContext` and the backend will recognize the tensor by their id 🤔

src/mlcontext.rs
```diff
@@ -0,0 +1,182 @@
...
+
+// types for MLContext
+
+// aka WebGpuDevice
+#[derive(Debug)]
+pub struct GpuDevice {}
```
theHamsta, line 90: we could use our own gpu type or have a `wgpu` feature to have this a `wgpu` GPU. `wgpu` could be used for render and compute on the Web as WebGPU or on desktop as (dx11, dx12, vulkan, metal, opengl)

matthargett: FWIW, I'm using wgpu in my WasmCore app runtime to present a WASI-gpu surface (and a bespoke WASI-xr surface). I'm letting MLProgram do the plan and runtime to spread the WebNN graph across CPU, GPU, and Neural Engine instead of manually pushing it to GPU. Maybe there's an interesting use case where ,for neural texture compression, we force it onto the GPU (depending on opcodes) so that the GPU tile's cache is already primed with the decoded texture?

Issue #101, opened 2026-04-22
CoreML MLProgram emitter: gaps surfaced by on-device watchOS validatio


matthargett:
## Background

I've been running rustnn's **CoreML MLProgram (MIL protobuf)** output on **Apple Watch SE 2 (arm64_32 ILP32, watchOS 11)**, end-to-end on the device itself — no host (macOS) compiler, no simulator, no developer tooling at inference time. Weights, graph, and all Apple intermediates live in ~10 MB of RSS and compile + execute within a jetsam-safe envelope. The motivation is a sovereign on-wrist WebNN runtime that can compile and run models shipped shared amongst peers. I've been chipping away at this for months without RustNN and have overcome many of the ILP32's CoreML compiler quirks to achieve the goal. In my patch stack, I now have all of the WPT WebNN tests passing on Apple Watch 6/SE2 running watchOS 11, alongside some larger integration examples from the community.

Full patch stack I've locally tested on real devices is here, let me know if you'd prefer them batched/grouped in some speicifc way: https://github.com/rebeckerspecialties/rustnn/pulls

### Stack
```
 webnn-graph JSON / programmatic GraphInfo
              │
              ▼
 rustnn  ──►  CoremlMlProgramConverter  ──►  MLProgram protobuf bytes
                                                │
                                                ▼
                                  MILTextCompiler (Swift, new)
                                   • protobuf → MIL text
                                   • MIL text → .mlmodelc bundle
                                   • streams to disk — jetsam-safe
                                                │
                                                ▼
                                      MLModel(contentsOf:)
                                       • Apple's CoreML loader
                                       • dispatches ANE / GPU / CPU
                                                │
                                                ▼
                                          on-device inference
```

`MILTextCompiler` is a small Swift package I built that turns rustnn's MLProgram protobuf into Apple's `.mlmodelc` bundle format *in-process on the watch* — normally this is a build-time step done by Xcode's `coremlc`. Apple ships no MLProgram compiler at runtime on watchOS, so we wrote one. I'm willing to open source it, if anyone has specific ideas on the best way to do that.

## What we've validated on-device so far

| Workload | rustnn ops exercised | On-watch result |
|---|---|---|
| LeNet-5 MNIST classifier (241 KB weights) | conv2d, add, relu, maxPool2d, reshape, matmul, softmax | **10/10** test digits classified correctly, all probabilities ≥ 0.92 |
[...]

PR #102, opened 2026-04-22
CoreML MLProgram: emit explicit gelu mode=EXACT

matthargett:
Refs #101.

## Problem

rustnn emits `gelu` MIL operations without a `mode` input, relying on the CoreML default. The iOS 17+ / macOS 14+ / macCatalyst MIL loaders are permissive and accept this, but the **watchOS 11 MIL loader** rejects the model during `MLModel(contentsOf:)` with:

```
Validation error parsing MIL model: Required param 'mode' is missing
```

This blocks every WebNN graph that contains a `gelu` op from loading on Apple Watch (Series 6+, SE 2, Ultra 1+, and later) running watchOS 11.

## Fix

WebNN's `gelu()` has no user-facing `mode` parameter; the spec defines the exact (erf-based) form as the implicit default, which maps to MIL's `mode = "EXACT"`. Emit this explicitly in the MLProgram protobuf. The change is a no-op on iOS/macOS loaders (same numerical behaviour, explicit vs. default), and unblocks `MLModel` loads on watchOS 11.

Adds a regression test that decodes the emitted MLProgram protobuf and asserts the `gelu` op carries an immediate `mode = string("EXACT")` input.

## How this was surfaced

Running a 2-layer char-level transformer (gelu-in-FFN) through rustnn → CoreML MLProgram → `MILTextCompiler` → Apple CoreML on **Apple Watch SE 2 (arm64_32, watchOS 11)** as part of an on-device WebNN conformance sweep. See #101 for the full context (stack diagram, on-device validation results, and the rest of the patch series).

## CI

All checks green on the fork PR ([rebeckerspecialties/rustnn#1](https://github.com/rebeckerspecialties/rustnn/pull/1)) against the current `main` (includes the recent `MLOperandDataType` / `MLOperatorOptions::label` / clippy-fix refactors): Rust Tests, RustNNPT Conformance Gate, Documentation Build all pass.

### DIFF
--- a/src/converters/coreml_mlprogram.rs
+++ b/src/converters/coreml_mlprogram.rs
@@ -2105,13 +2105,16 @@ impl CoremlMlProgramConverter {
                 // WebNN modes: "constant", "edge", "reflection", "symmetric"
             }
 
-            Operation::Gelu { .. }
-                // gelu: x (mode is optional, defaults to "EXACT")
-                if !input_names.is_empty() => {
+            Operation::Gelu { .. } => {
+                // gelu: x, mode
+                if !input_names.is_empty() {
                     inputs.insert("x".to_string(), Self::create_argument(&input_names[0]));
                 }
-                // CoreML GELU supports "EXACT" and "TANH_APPROXIMATION" modes
-                // WebNN GELU has no mode parameter (uses exact by default)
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

tarekziade:
Great, thanks a lot for this contribution!

PR #104, opened 2026-04-24
Add support for large ONNX files 

mtavenrath:
* Use shared webnn_graph resolve_external_weights function instead of duplicating the logic.
* Move ONNX weights from protobuf to external weights to support models >2gb
* Add shape_inference for GEMM

## Summary

## Validation

- [x] `make test`

## Documentation

### DIFF
--- a/src/converters/onnx.rs
+++ b/src/converters/onnx.rs
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-use crate::converters::{ConvertedGraph, operand_name};
+use crate::converters::{ConvertedGraph, ONNX_EXTERNAL_WEIGHTS_FILENAME, operand_name};
 use crate::debug_print;
 use crate::error::GraphError;
 use crate::graph::{DataType, Dimension, GraphInfo, OperandKind, get_static_or_max_size};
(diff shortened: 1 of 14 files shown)


src/converters/onnx.rs
```diff
@@ -1022,6 +1064,24 @@ impl OnnxConverter {
         GraphError::InvalidConversionOperand { operand }
     }
 
+    /// Pretty-print the webnn-graph-json style [`webnn_graph::ast::Node`] for `op_index` to stderr (ONNX conversion diagnostics).
+    fn eprintln_failed_operation_webnn_json(graph: &GraphInfo, op_index: usize) {
```
theHamsta, line 1068: why don't we make this a proper error? Ok for this review, but pls add a TODO.

We can stack errors with the thiserror crate we are using. We just shouldn't use GraphError for everything or add a `source` field to `GraphError`. thiserror will resolve a `source` field as a stacked error (e.g. GraphError caused by <underlying error>

other error prints should ideally be `error!` from log crate

src/loader.rs
```diff
@@ -63,11 +63,37 @@ pub fn sanitize_webnn_identifiers(text: &str) -> String {
     result
 }
 
+fn map_weight_resolve_error(err: WeightResolveError) -> GraphError {
```
theHamsta, line 66: same here, thiserror allows to automatically implement `from` and `source`. We should resolve this in a TODO later.

theHamsta (approved):

PR #111, opened 2026-04-29
feat: add MLContext, MLGraphBuilder API

theHamsta:
## Summary

MLContext accelerates inference by having persistent context, graph and tensor objects.
MLGraphBuilder is still only lightly tested.

## Validation

- [ ] `make test`

## Documentation


theHamsta:
This update ORT to rc12. Reasons why I think a new ORT would be useful:
- rc12 provides api22 which allows to query EP devices (what CPUs, NPUs, GPUs are available for which EP?). This feature is actually used here
- there seems to be api24 available, but not all new ORT features expose. New ORT features that could be useful
      - with dynamic EP registration from DLL
      - device streams to run inference
      - explicit device <-> host transfers
      - D3D12/Vulkan interop (for wgpu interop)
      - ORT values staying alive between multiple ort session

mtavenrath observed regression in CPU WPT test with this version bump.

src/mlgraphbuilder.rs
```diff
@@ -0,0 +1,2512 @@
...
+                }
+                op.kind = OperandKind::Output;
+                op.name = Some(name.to_string());
+            }
+            graph.output_operands.push(operand.id as u32);
+        }
+
+        debug!("Building graph with {} operands", graph.operands.len());
+        trace!("Building graph:\n{graph:#?}");
+        trace!("Graph webnn JSON:\n{}", {
```
theHamsta: both those trace calls, have the risk of being too verbose even for `trace!`. We should have a limit to do this only for small graphs

theHamsta: ideally, to reprs would omit weights and elide too long info in the graph.

Made this print only for graphs with < 20 operands.

Code is not executed if not `trace` log level enabled.

theHamsta:
Please note that this bumps the `ort` dependency. ONNX runtime 1.24.X will be needed as advertised by the `ort` crate

mtavenrath (approved):

PR #124, opened 2026-05-20
Add ResNet-50 example

mklimenko-nv:
A WIP PR to run an E2E model via rustnn. Currently depends on https://github.com/rustnn/webnn-graph/pull/16 for the ONNX model conversion to be merged.

After fixes related to marking all the nodes as outputs, perfomance of rustnn with TensorRT-RTX backend is similar to native execution.

To be added:
- [x] Full E2E run from ONNX
- [x] Performance comparison against TensorRT-RTX
- [ ] FP16 execution


mtavenrath:
@mklimenko-nv Can you rebase and cleanup the sample to use the new MLBuilder interface?

mklimenko-nv:
@mtavenrath, rebased and switched to the new MLBuilder interface, the sample is much cleaner now.

Cargo.toml
```diff
@@ -68,6 +68,7 @@ tempfile = "3.8"
 half = "2.4"
 safetensors = "0.7"
 insta = { version = "1.47.2", features = ["yaml"] }
+image = { version = "0.25", default-features = false, features = ["jpeg", "png"] }
```
mtavenrath, line 71: @theHamsta I am wondering if we can specify this dependency in a way that it's only for the examples or document that this dependency is required for the resnet example.

mtavenrath (approved):
This is a nice and clean example now. 

In the long term I am wondering if we want to stick with .webnn files for smaller samples or if it'd make sense to have a full builder example for smaller networks as well.

Cargo.toml
mklimenko-nv: @mtavenrath, it's located in `[dev-dependencies]`, so it should be used only for examples and tests, not when rustnn is added as a dependency. @theHamsta, could you please confirm this understanding?

mtavenrath: Thanks for clarifying. Merging.

PR #126, opened 2026-05-22
fix: validate input/output shapes for MLContext::dispatch

theHamsta:
## Summary

Asked cursor to validate the input/output shape before `MLContext::dispatch`. Not sure yet whether it did a good job. It used existing validation logic that I was not 100% aware of, but it went quite verbose.

I think this can be still improved. ~Especially since it somehow causes a test to fail and~ errors lack relevant contextual information (e.g. what operation/operand failed validation)

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -299,6 +299,21 @@ pub enum GraphError {
         expected: usize,
         actual: usize,
     },
+
+    #[error(
+        "graph input operand ids doesn't match the ids of OperandKind::Input: {input_ids:?} vs {input_ids_in_operands:?}"
+    )]
+    InputIdListMismatch {
+        input_ids: Vec<u32>,
+        input_ids_in_operands: Vec<u32>,
+    },
+    #[error(
+        "graph output operand ids doesn't match the ids of OperandKind::Input: {output_ids:?} vs {output_ids_in_operands:?}"
+    )]
+    OutputIdListMismatch {
+        output_ids: Vec<u32>,
+        output_ids_in_operands: Vec<u32>,
+    },
     #[error("graph input operand list does not match operand table")]
     InputOperandListMismatch,
     #[error("graph output operand list does not match operand table")]
--- a/src/error.rs
+++ b/src/error.rs
@@ -96,6 +96,15 @@ pub enum GraphBuilderError {
     #[error("Failed to build: requested MLGraphBuilder.build with an empty output map")]
     EmptyOutputHashMap,
 
+    #[error(
+        "Duplicates in outputs: used MLOperand {operand:?} for ouput name {first_name:?} and {second_name:?}"
+    )]
+    DuplicateOutput {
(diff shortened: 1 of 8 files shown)


src/mlcontext.rs
```diff
@@ -167,7 +168,101 @@ impl MLTensor {
...
+            .map(|(name, op)| (name.clone(), op.descriptor.clone()))
+            .collect()
+    }
+
+    fn verify_dispatch_bindings(
+        &mut self,
+        inputs: &HashMap<&str, &MLTensor>,
+        outputs: &HashMap<&str, &MLTensor>,
+    ) -> Result<()> {
+        let input_shapes: HashMap<String, Vec<usize>> = inputs
```
theHamsta, line 203: I went through `RuntimeShapeState`, I think we can get its logic localized here in a shorted code, but from the checks and error types it emits it looks fairly complete and ok. So, it should probably ok to merge this.

src/error.rs
```diff
@@ -299,6 +313,21 @@ pub enum GraphError {
...
+        input_ids: Vec<u32>,
+        input_ids_in_operands: Vec<u32>,
+    },
+    #[error(
+        "graph output operand ids doesn't match the ids of OperandKind::Input: {output_ids:?} vs {output_ids_in_operands:?}"
+    )]
+    OutputIdListMismatch {
+        output_ids: Vec<u32>,
+        output_ids_in_operands: Vec<u32>,
+    },
```
theHamsta, line 333: those are both internal errors and should eventually be removed once we have a `GraphInfo` that can't have invalid state

src/graph.rs
```diff
@@ -398,4 +480,109 @@ mod tests {
...
+    fn validate_io_operand_lists_accepts_consistent_graph() {
+        assert!(sample_io_graph().validate_io_operand_lists().is_ok());
+        assert!(sample_io_graph().io_binding_maps().is_ok());
+    }
+
+    #[test]
+    fn validate_io_operand_lists_rejects_extra_input_operand_id() {
+        let mut graph = sample_io_graph();
+        graph.input_operands.push(99);
+        std::assert_matches!(
```
theHamsta, line 525: Important! This bump minimum Rust version to the one from 28th May.

"assert!(matches!(" just returns "failed", `assert_matches!` return both left and right side in case of error, indicating how it failed and how to fix

src/error.rs
mtavenrath: What if we have a bug in our logic to handle the graph? Those errors would detect it.

mtavenrath (approved):

src/error.rs
theHamsta: GraphInfo should not allow to mess with its internals. It should only allow to mutate via methods that never destroy its invariance. If invariance of data structure is violated also panic instead of error might be appropiate

PR #127, opened 2026-05-28
deps: update trtx dependency to version 0.6.0+rtx1.5

theHamsta:
## Summary

Update trtx dependency to version `0.6.0` for support for TensorRT RTX 1.5.

 TensorRT RTX 1.5 helps us to pass more conformance tests.

## Validation

- [ ] `make test`

## Documentation


tests/wpt_conformance/mod.rs
```diff
@@ -826,12 +793,6 @@ pub fn run_all_trtx() -> Result<(), String> {
 
         for test_case in &file.tests {
             println!("  running: {} :: {}", op, test_case.name);
-            if let Some(reason) = trtx_skip_reason(test_case) {
-                skipped += 1;
-                println!("    [SKIP] {}", reason);
-                test_results.push(format!("{}    [SKIP] {}", test_case.name, reason));
-                continue;
-            }
```
theHamsta, line 834: don't skip, get notified on behavior changes of the failing tests (e.g. when they are fixed)

theHamsta: good idea on how to keep the skip reasons? Or is it obvious that they are all related to {u}int support?

mtavenrath: The skip reasons were all related to (u)int support. Clamp uint64 should still fail.

If we don't skip not yet supported tests we'll fail the CI. Can we have 'expected fails' which do not get counted as failure?

mtavenrath (dismissed):

tests/wpt_conformance/mod.rs
theHamsta: The skip reasons didn't represent current state of the passing tests. Some of the non-skipped tests were also failing.

We changed to a snapshot were every change in the behavior is a failure and must be updated (even a fail -> pass). @mklimenko-nv and me had the desire to describe the actual state of the converter

theHamsta: > If we don't skip not yet supported tests we'll fail the CI. Can we have 'expected fails' which do not get counted as failure?

yes, the snapshot has the concrete tests that are failing with their error reason. If the behavior changes, the snapshot needs to be updated. We checked that snapshot output is consistent between different GPUs and OSes

mtavenrath: In this case I do not have any objections. I'll merge the change.

PR #128, opened 2026-05-29
Caching for TRTX

theHamsta:
## Summary

Cache engines without weights, refit with current engine weights

Depends on #127 

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/backends/trtx.rs
+++ b/src/backends/trtx.rs
@@ -199,8 +199,9 @@ impl<'context> TrtxContext<'context> {
         let cuda_ctx = CudaContext::new(cuda_device_idx as usize)?;
         let mut builder = trtx::Builder::new(&LOGGER)?;
         let mut config = builder.create_config()?;
-        // Strip weights from engine and allow refitting
-        config.set_flag(trtx::trtx_sys::BuilderFlag::kREFIT);
+        // Strip marked weights weights from engine and makes them refittable, keeps other weights
+        // (mostly scalars)
+        config.set_flag(trtx::trtx_sys::BuilderFlag::kREFIT_INDIVIDUAL);
         config.set_flag(trtx::trtx_sys::BuilderFlag::kSTRIP_PLAN);
         let config = Rc::new(config.into());
         let runtime = Rc::new(trtx::Runtime::new(&LOGGER)?.into());
@@ -273,6 +274,9 @@ impl<'context, 'builder> MLBackendBuilder<'context, 'builder> for TrtxBuilder<'c
                 .take()
                 .expect("Frontend API should prevent TrtxBuilder::build to be called twice");
             crate::converters::TrtxConverter::build_network(&graph, &mut network)?;
+            for constant_id in graph.constant_operand_ids_to_handles.keys() {
+                network.mark_weights_refittable(&format!("{constant_id}"))?;
+            }
 
             let host_mem = self
                 .builder
@@ -326,7 +330,7 @@ impl<'context, 'builder> MLBackendBuilder<'context, 'builder> for TrtxBuilder<'c
                     }
                     .into());
                 }
-                let weight_name = format!("{id} CONSTANT");
+                let weight_name = format!("{id}");
                 trace!("Trying to refit weight {weight_name}");
                 unsafe {
                     refitter.set_named_weights_with_location(
--- a/src/backends/trtx.rs
+++ b/src/backends/trtx.rs
@@ -1,5 +1,8 @@
+use std::borrow::Cow;
 use std::collections::HashMap;
 use std::ffi::c_void;
+use std::ops::Deref;
+use std::sync::LazyLock;
 
 use cudarc::driver::CudaSlice;
 use cudarc::driver::CudaStream;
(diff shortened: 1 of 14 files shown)


theHamsta:
Do you want to have a way to disable this or have it disabled by default. It can be confusing when you try to modify the trtx converter when the converter is skipped entirely because of caching

src/converters/trtx.rs
```diff
@@ -535,11 +535,21 @@ impl TrtxConverter {
...
                 } else {
-                    network.add_constant(
+                    let mut layer = network.add_constant(
                         &add_dims,
                         data,
                         Self::webnn_to_trt_dtype(operand.descriptor.data_type)?,
-                    )
+                    );
+                    if let Ok(layer) = layer.as_mut() {
+                        //TODO: use API to name weights
```
theHamsta: needs to be added as a breaking change to trtx and RustNN's trtx converter. This is a work around that works for now without big code changes.

When you name the layer, the name of the constant will be `<constant layer name> CONSTANT`

theHamsta:
Need to think about constants that are not Constant for WebNN, e.g. small helper constants that the trtx converter created on its own. How to keep track of them, when weights get stripped?

theHamsta:
> Need to think about constants that are not Constant for WebNN, e.g. small helper constants that the trtx converter created on its own. How to keep track of them, when weights get stripped?

use kSTRIP_INDIVIDUAL + markWeighstRefitable to not strip all weights.

src/backends/trtx.rs
```diff
@@ -13,16 +16,57 @@ use std::rc::Rc;
...
+use crate::converters::TrtxConverter;
 use crate::error::Error;
 
+use crate::error::GraphBuilderError;
 use crate::mlcontext::MLTensor;
 use crate::mlcontext::{ListDevices, MLOperand};
 use crate::mlcontext::{MLBackendBuilder, MLGraph};
 use crate::mlcontext::{MLBackendContext, MLBackendGraph};
 
+// TODO: also used in trtexec-rs. Should be part of trtx API?
```
mtavenrath, line 36: It'd be reasonable to have this in trtx-rs, probably with a slightly more descriptive name.

if we'd ever want to introduce weight streaming, would it make sense to have a mapping interface instead so that trtx can request the data at the time it's required?

src/converters/trtx.rs
mtavenrath: adding some automatism to the weights name in layer.set_name seems reasonable as long as an already set weights name is not overwritten.

still, i am wondering why we wouldn't be able to call setWeightsName directly after creating the constant layer so that we do not have to change trtx-rs

theHamsta:
Ok, weight naming would be added to trtx here: https://github.com/rustnn/trtx-rs/pull/117

We can mark all weights that we name as refittable and then `kSTRIP_PLAN + kREFIT_INDIVIDUAL` will only strip the weights that we named.

src/backends/trtx.rs
theHamsta: See here https://github.com/rustnn/trtx-rs/pull/104 . Suggestion for name?

> if we'd ever want to introduce weight streaming, would it make sense to have a mapping interface instead so that trtx can request the data at the time it's required?

this does not hold any weights, since weights are stripped from plan. This is just the engine description (which topology and which kernels to use). TRT can do weight streaming from engine but since we get the weights from WebNN API we probably don't ever want them in the engine. Even if weights come from browser cache they will not be stored together with the TRT engine cache.

Since we use kREFIT_INDIVIDUAL without kREFIT_IDENTICAL we can have the same engine for graphs with different weights.  `kREFIT_IDENTICAL` allows constant folding, but would require us to hash the weights.

theHamsta: Also if we stay with TRT API: streaming weights from an engine file does not happen via IHostMemory but via IStreamReaderV2.

src/converters/trtx.rs
theHamsta: replaced this by naming the weights and not the layer.

>  as an already set weights name is not overwritten.

weights, aka constants are operands. Operands have no name in WebNN. Only Operations can have labels

theHamsta:
Depends on release of trtx with https://github.com/rustnn/trtx-rs/pull/117, otherwise ready. A test with a constant that is not a WebNN operand is missing.

src/converters/trtx.rs
```diff
@@ -9032,35 +9055,29 @@ impl TrtxConverter {
...
-                (None, None, conv_weights_owned)
-            } else {
-                (Some(filter_data), bias_raw, None)
-            }
+            let conv_weights_owned = OwnedConvWeights {
+                kernel: OwnedWeights {
+                    shape: conv_kernel_const_shape,
+                    data_type: TrtDataType::kFLOAT,
+                    values: kernel_values,
+                    name: None, // name added later
```
theHamsta: TODO: this not correct!

This is a transformed version of the WebNN weights, so we need to do the swizzle in TRT ops instead of transforming the weights or we need to repeat the transforms when refitting

Same also for cases where data type of WebNN weights are changed.

theHamsta: I tried to remove CPU transformations of Constants to easier refit and have weightless engines.

I did this automated with codex. Tests in this repo seem to pass still.

theHamsta:
I disabled the feature for now. It seems to work in the very limited test cases that we have for MLContext, but certainly require more test coverage to always enable. Also testing needs to consider the empty cache vs cache preset case. Caching might also be confusing when doing changes to trtx converter when cache is enabled.

mtavenrath (approved):

PR #131, opened 2026-06-03
Fix validation and shape inference 

mtavenrath:
for scatter_nd_shape, where, element_wise_logical and cast operators. Those failures have been uncovered during onnx2webnn conversion from smollm_135m.

## Summary

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/mlgraphbuilder.rs
+++ b/src/mlgraphbuilder.rs
@@ -22,9 +22,9 @@ use crate::shape_inference::{
     infer_conv_transpose2d_shape, infer_conv2d_shape, infer_expand_shape_dimensions,
     infer_gather_shape_dimensions, infer_gemm_shape_dimensions, infer_global_pool_shape,
     infer_matmul_shape_dimensions, infer_pad_shape, infer_pool2d_shape_dimensions,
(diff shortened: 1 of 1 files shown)


src/mlgraphbuilder.rs
```diff
@@ -818,6 +868,43 @@ fn same_shape(
...
+    inputs: &[MLOperand],
+    operation: &Operation,
+    graph: &GraphInfo,
+) -> Result<OperandDescriptor> {
+    let mut output = same_shape(inputs, operation, graph)?;
+    output.data_type = DataType::Uint8;
+    Ok(output)
+}
+
+fn unary_element_wise_logical_shape(
```
theHamsta, line 885: Not sure if I  would translate `element-wise` to `_elementwise_` instead of ` _element_wise_`

theHamsta (approved):

src/mlgraphbuilder.rs
theHamsta: would just one `elementwise_logical_shape` also do it? it can also handle the unary case

Issue #85, opened 2026-04-06
Add burn as backend


sagudev:
This is interesting because it is written in rust and it supports multiple backends (most interestingly wgpu).

mtavenrath:
Thanks for bringing this up. This project definitively looks interesting, especially given the WebGPU and Vulkan support enabling basically every non-coreML device of interest.

Topics to be evaluated
* Is there support for WebGPU subgroup matrices? 
* Is there support for Vulkan VK_KHR_cooperative_matrix and VK_NV_cooperative_matrix2?
* Does burn support the whole WebNN operator set? Likely given ONNX can be imported.
* Is everything JITted or are there AOT kernels as well? I assume one could generate AOT kernels

sagudev:
There is cooperative_matrix implementation in wgpu (when using core or what is know as native), but I am not sure burn uses it.

PR #135, opened 2026-06-04
POC for burn backend on the CPU and WebGPU. 

mtavenrath:
WPT and rustnnpt tests are successful, yet smollm_mlcontext does panic with an OOM error message during autotuning.

This POC is in its early stages and will requires a deep and thorough cleanup before starting the review process.

Fix issue #85 

## Summary

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/examples/smollm_mlcontext.rs
+++ b/examples/smollm_mlcontext.rs
@@ -1,8 +1,14 @@
+//! SmolLM generation via MLContext (dynamic KV-cache shapes).
+//!
+//! Build with `dynamic-inputs` (required) and a runtime, e.g.:
+//! `cargo run --example smollm_mlcontext --features dynamic-inputs,burn-runtime-cpu -- --backend burn-cpu`
 use std::{collections::HashMap, io::Write, path::PathBuf, process::Command};
 
 use anyhow::{Context, anyhow, bail};
-use clap::Parser;
+use clap::{Parser, ValueEnum};
 use log::{debug, info};
+#[cfg(any(feature = "burn-runtime-cpu", feature = "burn-runtime-webgpu"))]
+use rustnn::BackendPreference;
 use rustnn::{
     ContextProperties, GraphValidator, ValidationArtifacts, load_graph_from_path,
     mlcontext::{
@@ -12,6 +18,17 @@ use rustnn::{
     operator_enums::MLOperandDataType,
 };
 use tokenizers::Tokenizer;
+
+#[derive(Debug, Clone, Copy, ValueEnum, Default)]
+enum BackendArg {
+    /// Use `accelerated` / `powerPreference` autoselection (TensorRT, ONNX, etc.).
+    #[default]
+    Auto,
+    #[cfg(feature = "burn-runtime-cpu")]
+    BurnCpu,
+    #[cfg(feature = "burn-runtime-webgpu")]
+    BurnWebGpu,
+}
 
 #[derive(Parser, Debug)]
 #[command(about = "Pure Rust SmolLM generation from .webnn + tokenizer.json")]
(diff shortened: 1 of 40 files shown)


src/mlcontext.rs
```diff
@@ -358,6 +306,29 @@ impl MLContextOptions {
     pub fn set_accelerated(&mut self, accelerated: bool) {
         self.accelerated = accelerated;
     }
+
+    /// Pin a specific backend instead of using autoselection.
+    #[cfg(any(feature = "burn-runtime-cpu", feature = "burn-runtime-webgpu"))]
+    pub fn with_backend_preference(mut self, preference: BackendPreference) -> Self {
```
theHamsta, line 312: any reason to not allow to set the device directly? Or is the idea that one chooses the backend without caring about the device?

My idea was that the user would first ask to list all available devices to also know that they are really available before they choose. The approach to set a preference and then silently fallback to someone else is probably more in line with Web APIs that try to not expose too much info about the users devices

src/backend_selection.rs
```diff
@@ -188,10 +213,73 @@ pub(crate) fn select_backend(options: &MLContextOptions) -> Result<BackendDevice
...
+        (_, true) if have_burn_webgpu => BackendDevice::BurnWebGpu,
+        #[cfg(feature = "burn-runtime-cpu")]
+        (_, false) if have_burn_cpu => BackendDevice::BurnCpu,
+        #[cfg(all(feature = "burn-runtime-cpu", not(feature = "burn-runtime-webgpu")))]
+        (_, true) if have_burn_cpu => BackendDevice::BurnCpu,
         _ => return Err(crate::error::Error::NoBackendAvialable),
     })
 }
 
+fn validate_backend_hint(hint: BackendDevice) -> Result<BackendDevice> {
```
theHamsta, line 226: the names of the functions should probably distinguish a backend_hint and a device_hint

examples/smollm_mlcontext.rs
```diff
@@ -1,8 +1,14 @@
+//! SmolLM generation via MLContext (dynamic KV-cache shapes).
+//!
+//! Build with `dynamic-inputs` (required) and a runtime, e.g.:
+//! `cargo run --example smollm_mlcontext --features dynamic-inputs,burn-runtime-cpu -- --backend burn-cpu`
```
theHamsta, line 4: this likely also works also in the browser (needs emscripten to compile tokenizers), burn can run webgpu or wasm32 in wasm32 mode

examples/smollm_mlcontext.rs
```diff
@@ -505,7 +525,20 @@ fn main() -> anyhow::Result<()> {
...
+    match args.backend {
+        BackendArg::Auto => {}
+        #[cfg(feature = "burn-runtime-cpu")]
+        BackendArg::BurnCpu => {
+            context_options = context_options.with_backend_preference(BackendPreference::BurnCpu);
+        }
+        #[cfg(feature = "burn-runtime-webgpu")]
+        BackendArg::BurnWebGpu => {
+            context_options =
+                context_options.with_backend_preference(BackendPreference::BurnWebGpu);
```
theHamsta, line 538: we need to decide on how to mark non-standard methods and data structures in the MLContext frontend. I prefixed until now with `rustnn_` all methods that were invention from us and not in WebGPU standard. Not that I say that is the best strategy, but we should make it consistent. Maybe just a doc marker enough

src/mlcontext.rs
mtavenrath: the reason is only it's a proof-of-concept.

On the desktop side being able to enumerate all devices and choose one a valid option. In the web my assumption is that the browser does the filtering. More feedback is required.

examples/smollm_mlcontext.rs
mtavenrath: I agree. A rustnn_ prefix is a good candidate to mark those methods.

PR #136, opened 2026-06-10
Weights but different

theHamsta:
## Summary

I would like to separate the RAM representation of weights from GraphInfo. How GraphInfo works today it is assumed that it has access to the weights, so I'm trying to do this in steps.

What I would like to support is to get the concrete values of weights as lazy as possible and possibly owned by an external object owned by the user. I created the trait `WeightContext` for users to support their object as a weight provider.

- GraphInfo + OwnedData (what we have today)
- GraphInfo + StringRef: we could implement WeightContext for `onnx::Graph` or a safetensors file
- GraphInfo + FileRef: e.g. ONNX/WebNN file + external weights
- GraphInfo + IdRef: e.g. MLContext with tensor ids, weights owned by the backend (e.g. already uploaded to accelerator)
- GraphInfo + Missing: explicitly work with graphs where weights are unknown
- GraphInfo + Zeroed/ZeroedUniquePtr: don't allocate memory for large memory that contain zeros, support weightless building in TRT

## Validation

- [ ] `make test`

## Documentation


src/graph.rs
```diff
@@ -198,14 +375,68 @@ pub struct GraphInfo {
     #[serde(default)]
     pub operations: Vec<Operation>,
     #[serde(default)]
-    pub constant_operand_ids_to_handles: HashMap<u32, ConstantData>,
+    pub constant_operand_ids_to_handles: HashMap<u32, ConstantReference>,
```
theHamsta, line 662: I really want to make all those members in steps `private` or `pub(crate)`. making `constant_operand_ids_to_handles` private would require a constructor for GraphInfo and more methods to manipulate

src/graph.rs
```diff
@@ -179,16 +262,110 @@ pub struct Operand {
...
+    pub url: String,
+    pub offset: u64,
+    pub len: u64,
+    #[serde(default)]
+    pub label: Option<String>,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ConstantData {
-    #[serde_as(as = "Base64")]
```
theHamsta, line 272: why is Base64 needed here? serde_json will do base64 automatically, which other serializer does not use base64 but base64 is desired? Where is GraphInfo serialized, for Firefox IPC?

theHamsta:
Need to combine with #128 and try with zeroed TRTX weights.

Issue #134, opened 2026-06-04
Add support for CoreML in new WebNN IDL API


mtavenrath:
RustNN now supports a rustified version of the WebNN Graphbuilder API. This API unifies access creation and execution across the various backends we have so far. Support for ONNX and trtx-rs has already been added, yet CoreML support is lacking.

Once CoreML support has been added other language integrations like node.js and PyWebNN can switch over to this interface to avoid the code duplication we currently by implementing the same WebNN API interface across all of the bindings.

theHamsta:
Fixed via #140

PR #140, opened 2026-06-14
feat(coreml): wire CoreML into the unified WebNN IDL API (#134)

tarekziade:
Add CoreML support to the rustified MLContext/MLGraphBuilder API, alongside the existing ONNX and trtx backends.

Known converter gaps (tracked separately): the CoreML MLProgram converter cannot yet load a graph whose output is a bare `identity` op, nor bind a declared-but-unused input. The unused_incompatible_inputs test skips CoreML execution of that degenerate graph; builder-level assertions still run on all backends.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,8 @@
 name = "rustnn"
 version = "0.5.11"
 edition = "2024"
+# MSRV: std::assert_matches! (used in tests) was stabilized in Rust 1.96.0.
+rust-version = "1.96"
 description = "W3C WebNN implementation with ONNX, CoreML, and TensorRT backends [DO NOT USE IN PRODUCTION - Development Release]"
 license = "Apache-2.0"
 repository = "https://github.com/rustnn/rustnn"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,8 +2,8 @@
 name = "rustnn"
 version = "0.5.11"
 edition = "2024"
-# MSRV: std::assert_matches! (used in tests) was stabilized in Rust 1.96.0.
-rust-version = "1.96"
+# Tests use std::assert_matches!, which requires Rust 1.96.0.
+rust-version = "1.95"
 description = "W3C WebNN implementation with ONNX, CoreML, and TensorRT backends [DO NOT USE IN PRODUCTION - Development Release]"
 license = "Apache-2.0"
 repository = "https://github.com/rustnn/rustnn"
--- a/src/backends/coreml.rs
+++ b/src/backends/coreml.rs
@@ -84,7 +84,7 @@ pub(crate) struct CoremlContext {
 }
 
 impl CoremlContext {
-    pub(crate) fn new(device_type: DeviceType) -> crate::error::Result<Self> {
+    pub(crate) fn new_from_device_type(device_type: DeviceType) -> crate::error::Result<Self> {
         Ok(Self {
             device_type,
             tensors: Vec::new(),
--- a/src/backends/coreml.rs
+++ b/src/backends/coreml.rs
@@ -1,0 +1,428 @@
+//! CoreML backend for the unified WebNN IDL API (macOS only).
+//!
+//! Mirrors the ONNX Runtime backend in `src/backends/ort.rs`: a [`CoremlContext`]
+//! owns raw-byte host tensor storage, a [`CoremlBuilder`] converts a [`GraphInfo`]
+//! to a CoreML MLProgram and compiles it once, and [`CoremlGraph`] holds the
+//! compiled model for repeated dispatch.
+
+#![cfg(all(target_os = "macos", feature = "coreml-runtime"))]
+
+use std::collections::HashMap;
+use std::fmt;
+
+use log::debug;
+
+use crate::GraphInfo;
+use crate::backend_selection::DeviceType;
+use crate::converters::{CoremlMlProgramConverter, GraphConverter};
+use crate::error::Error;
+use crate::executors::coreml::{
+    CompiledCoremlModel, CoremlByteInput, compile_model, run_coreml_bytes,
+};
+use crate::mlcontext::{
+    MLBackendBuilder, MLBackendContext, MLBackendGraph, MLGraph, MLTensor, MLTensorDescriptor,
+};
+
+/// Number of bytes required to store a tensor described by `descriptor`.
+fn tensor_byte_len(descriptor: &MLTensorDescriptor) -> usize {
+    descriptor.rustnn_required_bytes()
+}
+
+/// Host tensor storage for the CoreML backend (mirrors `OrtTensor`).
+#[derive(Debug)]
+pub(crate) struct CoremlTensor {
+    memory: Vec<u8>,
+}
+
+/// A compiled CoreML model held by [`MLGraph`] (mirrors `OrtGraph`).
+pub(crate) struct CoremlGraph {
+    model: CompiledCoremlModel,
+}
+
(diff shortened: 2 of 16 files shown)


src/graph.rs
```diff
@@ -574,30 +575,30 @@ mod tests {
...
+        ));
     }
 
     #[test]
     fn validate_io_operand_lists_rejects_output_kind_not_in_list() {
         let mut graph = sample_io_graph();
         graph.operands[0].kind = OperandKind::Output;
         graph.output_operands.push(0);
-        std::assert_matches!(
+        assert!(matches!(
```
theHamsta: std::assert_matches was just added with the most recent stable version.

It would be good to have a MSRV policy defined somewhere for the library itself and for dev-code (samples and tests).

`std::assert_matches!` will show both sides of the match in case of error while `assert!(matches!` will just show `assert` failed.

src/mlcontext.rs
```diff
@@ -474,7 +485,19 @@ impl<'context> MLContext<'context> {
...
+                {
+                    Box::new(crate::backends::coreml::CoremlContext::new(device_type)?)
+                }
+                #[cfg(not(all(target_os = "macos", feature = "coreml-runtime")))]
+                {
+                    let _ = device_type;
+                    // `select_backend` only yields `Coreml` when the feature and platform are
+                    // available, so this arm is unreachable in practice.
+                    unreachable!("CoreML backend selected without coreml-runtime support")
+                }
```
theHamsta: for OnnxContext and TrtxContext, I just mapped them to `DisabledContext` when the features are disabled, so users can always use the types (but get an error when they use them without the features)

https://github.com/rustnn/rustnn/blob/5847703ca5eec69f7c8cd4be1ccba27a612d6234/src/backends/mod.rs#L83-L104

just to reduce the conditional code a bit in situations where you are dealing with all of the backends

src/backends/coreml.rs
```diff
@@ -0,0 +1,428 @@
...
+
+/// Number of bytes required to store a tensor described by `descriptor`.
+fn tensor_byte_len(descriptor: &MLTensorDescriptor) -> usize {
+    descriptor.rustnn_required_bytes()
+}
+
+/// Host tensor storage for the CoreML backend (mirrors `OrtTensor`).
+#[derive(Debug)]
+pub(crate) struct CoremlTensor {
+    memory: Vec<u8>,
```
theHamsta, line 34: would it be more efficient to make this a MLTensor/MLMultiArray in a future change? Or is this mostly irrelevant since MacOS uses unified memory and those types are just zero-copy wrappers around the raw memory

I would have loved to use device memory also with the ORT backend, but unfortunately the `ort` crate doesn't the newest revision of the ORT. So I mostly made it just functional and not yet optimized.

Features that I could use for trtx but not for ORT (onnxruntime has all those features in the newest version, but not in the `ort` crate):
- compile API
- device memory
- explicit host<-> device transfers
- explicit device synchronization
- graphics interop API (also not used for trtx, and not needed at the moment as there is no WebGPU<->WebNN interop yet apart from device creation)

src/executors/coreml.rs
```diff
@@ -123,6 +123,311 @@ pub fn run_coreml_with_inputs_checked(
...
+            return Ok(CompiledCoremlModel {
+                model,
+                compute_unit: name,
+                compiled_dir,
+                temp_model,
+            });
+        }
+
+        // No compute unit could load the model: clean up the on-disk artifacts.
+        let _ = std::fs::remove_dir_all(&compiled_dir);
```
theHamsta, line 233: could helpers in `crate::backends::caching` be useful to potentially skip conversion https://github.com/rustnn/rustnn/pull/128 ?

theHamsta (dismissed):
LGTM!

What were your experiences with the abstractions that I created very adhoc? We should change anything that you feel could be done better or does not fit well with coreml.

src/executors/coreml.rs
theHamsta: but trtx uses caching without the weights, to just cache the compiled network topology

tarekziade: sounds like a good improvement, I would do it as a follow up

theHamsta (dismissed):

tarekziade:
> LGTM!
> 
> What were your experiences with the abstractions that I created very adhoc? We should change anything that you feel could be done better or does not fit well with coreml.

Thanks! The abstractions held up well. I wired CoreML through MLBackendContext/MLBackendBuilder/MLBackendGraph without fighting them. Two things I bumped into:

1. The one worth revisiting is the byte-oriented tensor API. Since read_tensor/write_tensor/dispatch deal in &[u8], my CoremlTensor is just a Vec<u8>, so dispatch copies bytes in, runs, and copies back out;  two copies I could skip by binding an MLMultiArray onto the tensor's memory (basically free on unified memory). Same tension you mentioned for ORT device memory. A per-backend tensor handle instead of mandated host bytes would get us both there. Good follow-up candidate.

2. Smaller: construction isn't in any trait, so the shared DisabledContext made me name CoreML's constructor new_from_device_type to dodge a duplicate-new collision with Trtx when both features are off. Works, but a tiny construction trait or per-backend Disabled structs would clean it up.

Cargo.toml
```diff
@@ -2,6 +2,8 @@
 name = "rustnn"
 version = "0.5.11"
 edition = "2024"
+# MSRV: std::assert_matches! (used in tests) was stabilized in Rust 1.96.0.
+rust-version = "1.96"
```
theHamsta: I believe consumers can even use a lower Rust version as they don't compile the tests. I thought it would be ok to use the new feature for tests as it does not affect consumers.

@shubhamg13 would this MSRV be a problem for servo?

shubhamg13: As of now we are still at `1.95.0` https://github.com/servo/servo/blob/main/rust-toolchain.toml

But I guess, It is okay since I can do my experiments without this MSRV change on my local.

shubhamg13 (approved):

theHamsta:
> > LGTM!
> > What were your experiences with the abstractions that I created very adhoc? We should change anything that you feel could be done better or does not fit well with coreml.
> 
> Thanks! The abstractions held up well. I wired CoreML through MLBackendContext/MLBackendBuilder/MLBackendGraph without fighting them. Two things I bumped into:
> 
>     1. The one worth revisiting is the byte-oriented tensor API. Since read_tensor/write_tensor/dispatch deal in &[u8], my CoremlTensor is just a Vec, so dispatch copies bytes in, runs, and copies back out;  two copies I could skip by binding an MLMultiArray onto the tensor's memory (basically free on unified memory). Same tension you mentioned for ORT device memory. A per-backend tensor handle instead of mandated host bytes would get us both there. Good follow-up candidate.
> 
>     2. Smaller: construction isn't in any trait, so the shared DisabledContext made me name CoreML's constructor new_from_device_type to dodge a duplicate-new collision with Trtx when both features are off. Works, but a tiny construction trait or per-backend Disabled structs would clean it up.

The host bytes were basically coming from coping the browser API where you provide JS host data to read/write tensors. Those methods imply some kind of copy from their semantics.

Do you imagine a method where a tensor could also be created from a borrow from host data? Or just allowing to optionally get a `mut &[T]`/`&[T]` onto the tensors owned data? I agree that this can be improved.

`&[u8]` was basically just done because I thought it could make backend's life easier by casting &[T] to &[u8] just in one place. u8+DataType enum is just easier than trying to find a Rust equivalent for every present and future ML data type that might exist (e.g. users might just load the bytes from `int4`/`mxfp4` without Rust supporting the types).

PR #145, opened 2026-06-17
Add `Send + Sync` bounds to the `MLBackendContext` and `MLBackendBuilder`

shubhamg13:
## Summary

Embeddings like [Servo](https://github.com/servo/servo) use multiple script threads dispatching WebNN graphs concurrently. Without `Send + Sync`, `MLContext` cannot be stored in a `Mutex` or shared across threads, forcing single-threaded use only.

## Validation

- [x] `make test`

## Documentation

### DIFF
--- a/src/backends/trtx.rs
+++ b/src/backends/trtx.rs
@@ -16,7 +16,6 @@ use std::sync::{Arc, Mutex};
 use trtx::CudaEngine;
 use trtx::ExecutionContext;
 use trtx::Refitter;
-use trtx::Tensor;
 use trtx::host_memory::HostMemory;
 
 use crate::GraphInfo;
@@ -218,13 +217,12 @@ impl<'context> TrtxContext<'context> {
 
 #[allow(dead_code)]
 pub(crate) struct TrtxBuilder<'builder> {
-    network: Option<trtx::NetworkDefinition<'builder>>,
+    network: Mutex<Option<trtx::NetworkDefinition<'builder>>>,
     builder: Arc<Mutex<trtx::Builder<'builder>>>,
     config: Arc<Mutex<trtx::BuilderConfig<'builder>>>,
     cuda_context: Arc<CudaContext>,
     runtime: Arc<Mutex<trtx::Runtime<'builder>>>,
     operands: HashMap<String, MLOperand>,
-    tensors: Vec<Tensor<'builder>>,
     strings: Vec<String>, //_parser: Option<OnnxParser<'builder>>,
     caching_enabled: bool,
 }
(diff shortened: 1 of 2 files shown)


shubhamg13:
cc: @xiaochengh

theHamsta:
I would need to add unsafe `Send + Sync` bounds to some of the `trtx` objects. I tried to make them multi-threading safe, but avoided to set the unsafe bounds prematurely. at least Send should be safe.

I try to fix this.

Could you change the `Rc`s in the trtx backend by `Arc`s? I used `Rc` because `Send` was was not required until now

theHamsta:
I also saw that some of the errors that can be thrown by the `ort` crate are not `Send+Sync`, so `Send+Sync` bounds are also missing on some of the sources of errors. So, errors need to be `.map_err` in a multi-threaded scenario which is also on my TODO list.

shubhamg13:
> I would need to add unsafe `Send + Sync` bounds to some of the `trtx` objects. I tried to make them multi-threading safe, but avoided to set the unsafe bounds prematurely. at least Send should be safe.
> 
> I try to fix this.
> 
> Could you change the `Rc`s in the trtx backend by `Arc`s? I used `Rc` because `Send` was was not required until now

I created a separate PR for same. Please take a look #147

shubhamg13:
I think we need these lines back https://github.com/rustnn/rustnn/pull/147/changes/732e6a6b51228780d237ea807aaf090d9031bcc5#diff-63e5d556d57db956975f8279cc2a6123d9770d0c72b7527ddb18fb90dcb131ffL183

theHamsta:
> I think we need these lines back [732e6a6#diff-63e5d556d57db956975f8279cc2a6123d9770d0c72b7527ddb18fb90dcb131ffL183](https://github.com/rustnn/rustnn/commit/732e6a6b51228780d237ea807aaf090d9031bcc5#diff-63e5d556d57db956975f8279cc2a6123d9770d0c72b7527ddb18fb90dcb131ffL183)

But trtx objects are not internally thread safe, at least `NetworkDefinition` is not. It should be possible just with safe Rust to use trtx. Maybe I need to revisit that traits on `trtx::Tensor` (I get an error even after adding a Mutex around NetworkDefinition).

I tested with only removing the unsafe trait impl for TrtxContext, because I didn't see the other one.

This needs to be fixed in `trtx` if there is still a problem.

EDIT: I think the problem is `trtx::Tensor` and other objects can hold phantomdata `NetworkDefintion`. Will fix

theHamsta:
> I think we need these lines back [732e6a6#diff-63e5d556d57db956975f8279cc2a6123d9770d0c72b7527ddb18fb90dcb131ffL183](https://github.com/rustnn/rustnn/commit/732e6a6b51228780d237ea807aaf090d9031bcc5#diff-63e5d556d57db956975f8279cc2a6123d9770d0c72b7527ddb18fb90dcb131ffL183)

You are right that with RustNN usage, we would only use the `network` on build which requires an exclusive reference. So it would be safe. I would prefer if consumers can make trtx work without manual `Send+Sync` impls.

theHamsta:
A Mutex and removing an unused `Vec<trtx::Tensor>` did the job

theHamsta:
Seems to be also compatible with #140 . Ship it!

PR #148, opened 2026-06-17
Add INT4/UINT4 support

mtavenrath:
## Summary

There are a few 4-bit integer WPT-webnn tests even though those datatypes have not yet added to the spec language. Add support for int4 and wire it up in the backends.

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/backends/ort.rs
+++ b/src/backends/ort.rs
@@ -16,6 +16,7 @@ use crate::backend_selection::BackendDevice;
 use crate::converters::{GraphConverter, OnnxConverter};
 use crate::error::Error;
 use crate::executors::onnx::ensure_ort_initialized;
+use crate::graph::{pack_int4, pack_uint4_from_i32, unpack_int4, unpack_uint4};
 use crate::mlcontext::{
     ListDevices, MLBackendBuilder, MLBackendContext, MLGraph, MLTensor, MLTensorDescriptor,
 };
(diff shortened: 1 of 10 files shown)


src/backends/ort.rs
```diff
@@ -189,8 +190,30 @@ fn session_input_from_host(
         return Ok(SessionInputValue::from(dyn_val));
     }
 
+    let shape_i64: Vec<i64> = shape.iter().map(|&d| d as i64).collect();
     let expected = elements.saturating_mul(elem_sz);
+    let host_packed = descriptor.data_type().rustnn_storage_byte_length(elements);
     if bytes.len() != expected {
+        if *ty == TensorElementType::Int32
+            && descriptor.data_type() == crate::operator_enums::MLOperandDataType::Int4
+            && bytes.len() == host_packed
```
theHamsta, line 199: ultimately, packing/unpacking should not be necessary (unless of course like here when the user really provides Int32 data.

Apart from creating a Value or ValueRef from untyped data, one could also create an own int4/uint4 (or int4x2 type) where people can interact with int4 data from the host side.
https://docs.rs/ort/2.0.0-rc.12/ort/value/type.Tensor.html
https://docs.rs/ort/2.0.0-rc.12/ort/value/trait.IntoTensorElementType.html

src/backends/ort.rs
```diff
@@ -450,6 +472,66 @@ fn copy_dyn_value_to_buffer(
...
+                return Err(Error::GraphDispatchError {
+                    source: format!(
+                        "output '{name}': packed int4 {} bytes, descriptor expects {}",
+                        packed.len(),
+                        expected
+                    )
+                    .into(),
+                });
+            }
+            dst.copy_from_slice(&packed);
```
theHamsta, line 503: all theses copies are still here, because ORT didn't provide the EP memory allocators back then (they had their own allocators for CUDA devices though, but now they have EP in their git version, which should allow direct copies to allocator or zero-copy inference)

theHamsta (approved):

PR #149, opened 2026-06-18
feat(litert): wire `LiteRT` backend

shubhamg13:
## Summary

- Add `schema.fbs` for `LiteRtConverter`
- Wire LiteRT backend module that implements the MLBackendContext / MLBackendBuilder traits.
- Runs WPT Conformance Tests with 19% passing rate
- Add some unit tests and integration test for converters/execution 

## Validation

- [x] `make test`

## Documentation


shubhamg13:
cc: @xiaochengh

shubhamg13:
This change is ready for review. Please take a look and let me know your thoughts.

build.rs
```diff
@@ -20,17 +22,38 @@ fn recurse(dir: &Path, files: &mut Vec<String>) {
...
-    // Only compile CoreML protos - ONNX protos come from webnn-onnx-utils
     config.compile_protos(&coreml_files, &[coreml_dir])?;
+    Ok(())
+}
+
+fn build_tflite_schema() {
+    let out_dir = env::var("OUT_DIR").unwrap();
+    if let Err(_) = run_flatc("protos/tflite/schema.fbs", &out_dir) {
+        println!("cargo:rerun-if-changed=protos/tflite/schema.fbs");
+    }
```
theHamsta: I think we should probably fail if `run_flatc` fails (with helpful message.

If flatc fails I get instead
```
error: couldn't read `/media/other_linux/data/projects2/rustnn/target/debug/build/rustnn-febf79af2f37906f/out/schema_generated.rs`: No such file or directory (os error 2)
 --> src/converters/litert.rs:7:1
  |
7 | include!(concat!(env!("OUT_DIR"), "/schema_generated.rs"));
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Also, ` println!("cargo:rerun-if-changed=protos/tflite/schema.fbs");` should happen in the success case not the error case, right?

I think also warnings in generated file need to be silenced. the flatc does not seem to be prepared for Rust edition 2024, so would probably need to suppress warnings in the file or on inclusion of this file

build.rs
```diff
@@ -20,17 +22,38 @@ fn recurse(dir: &Path, files: &mut Vec<String>) {
...
+fn build_coreml_protos() -> Result<(), Box<dyn std::error::Error>> {
     let mut config = prost_build::Config::new();
     config.bytes(["."]); // Fix clippy::needless_borrows_for_generic_args
 
     let coreml_dir = "protos/coreml";
-
     let mut coreml_files = collect_protos(coreml_dir);
     coreml_files.sort();
-
-    // Only compile CoreML protos - ONNX protos come from webnn-onnx-utils
```
theHamsta, line 32: is removing this comment related to this change?

Makefile
```diff
@@ -243,6 +247,9 @@ help:
 	@echo "  coreml             - Convert graph to CoreML format"
 	@echo "  coreml-validate    - Convert and validate CoreML graph"
 	@echo ""
+	@echo "LiteRT Conversion:"
+	@echo "  litert             - Convert graph to LiteRT/TFLite format"
+	@echo ""
 	@echo "Documentation:"
```
theHamsta, line 287: could we add some docs where users can get `libLiteRt.so` or the likes for their platform? onnxruntime also does not really have that documented, but provides a makefile recipe to download the runtime.

We can also move that to a follow-up and document getting the runtime libs for all the backends.

theHamsta: This PR seems to link litert. Is dynamic loading on demand possible like with onnxruntime/trtx?

theHamsta: litert 0.3 says it would download the binaries https://github.com/offbit-ai/LiteRT#environment-variables-escape-hatches

it did not do that for me on Linux with 0.2.1

theHamsta: Ok, I have it under
```
$HOME/.cache/litert-sys/v0.10.2/x86_64-unknown-linux-gnu/
```
when I build litert-rs. For it to work, it would probably be necessary to set `rpath` (as the README explains for macos) or for it to work or use dynamic linking. Settings LD_LIBRARY_PATH to that also solves that for me.

theHamsta: I'm getting a SEGV for concurrent execution of the tests on Linux

Codex thinks it has isolated the issue to concurrently running those tests and that they run fine in isolation

```
  Minimal reproducer:

  target/debug/deps/rustnn-ff98150625141aa4 mlgraphbuilder::test \
    --skip mlgraphbuilder::test::add_mat_plus_scalar \
    --skip mlgraphbuilder::test::quantize_dequantize_linear_output_dtype

  That runs only:

  - mlgraphbuilder::test::add_inputs
  - mlgraphbuilder::test::unused_incompatible_inputs

  and exits 139.

  So the crashing interaction is mlgraphbuilder::test::unused_incompatible_inputs running concurrently with mlgraphbuilder::test::add_inputs. Individually,
  unused_incompatible_inputs does not SEGV; it fails its assertion at src/mlgraphbuilder.rs:3083.
```

crash stack trace is related to a free
```
#0  0x000075cb5308abe0 in ?? ()
#1  0x000075cb53ea10d2 in __GI___nptl_deallocate_tsd () at ./nptl/nptl_deallocate_tsd.c:73
#2  __GI___nptl_deallocate_tsd () at ./nptl/nptl_deallocate_tsd.c:22
#3  0x000075cb53ea3f70 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:462
#4  0x000075cb53f377ac in __GI___clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```
I can't guarantee that I have the correct runtime version.

I can run single-threaded with
```
cargo t -F litert-runtime  -- --test-threads 1
```

src/backend_selection.rs
```diff
@@ -129,7 +129,7 @@ pub(crate) fn select_backend(options: &MLContextOptions) -> Result<BackendDevice
     #[cfg(feature = "litert-runtime")]
     let have_litert = cfg!(feature = "litert-runtime");
     #[cfg(feature = "litert-runtime")]
-    let want_litert = false;
+    let want_litert = true;
```
theHamsta: could you update this to
```
 let want_litert =
        options.backend_hint.is_none() || options.backend_hint == Some(Backend::LiteRt);
```

I have updated the code to introduce backend hints.

.github/workflows/ci.yml
```diff
@@ -33,6 +33,13 @@ jobs:
...
           protoc --version
 
+      - name: Install flatc
+        run: |
+          curl -sSL "https://github.com/google/flatbuffers/releases/download/v25.12.19/Linux.flatc.binary.clang++-18.zip" -o /tmp/flatc.zip
+          sudo unzip -o /tmp/flatc.zip flatc -d /usr/local/bin
+          sudo chmod +x /usr/local/bin/flatc
+          flatc --version
+
       - name: Cache cargo registry
```
theHamsta, line 43: would it be possible to run (some) tests on CI? it should be able to run on CPU. Can also be a follow-up

This way we don't accidentally break your code.

src/backends/litert.rs
```diff
@@ -1,18 +1,27 @@
 use std::collections::HashMap;
```
theHamsta, line 1: it would be nice if we could set the logging to use the log crate

litert-rs looks for C symbol `LiteRtGetDefaultLogger`

https://github.com/offbit-ai/LiteRT/blob/607533e9fc3fd79bb68459aeadbcc602aab2ff9f/litert/src/logging.rs#L108-L112

```
//! Controls for LiteRT's internal logging.
//!
//! LiteRT emits `INFO` / `WARNING` messages during environment initialisation
//! (accelerator discovery, XNNPACK setup, etc.). Apps embedding LiteRT often
//! want to quieten this output. Use [`set_global_log_severity`] once at
//! startup to filter messages globally.
//!
//! ## Availability
//!
//! The logger-control symbols (`LiteRtGetDefaultLogger`,
//! `LiteRtSetMinLoggerSeverity`) are exported by the macOS build of
//! `libLiteRt` but **not** by the Linux, Windows, or Android Maven builds we
//! consume today. On platforms that lack these symbols,
//! [`set_global_log_severity`] returns [`Error::Unsupported`] rather than a
//! link error, and LiteRT continues to log at its default verbosity.
```
I assume we could provide it, instead of it logging to stdout without filtering support. Maybe also fixable upstream

Makefile
shubhamg13: > litert 0.3 says it would download the binaries https://github.com/offbit-ai/LiteRT#environment-variables-escape-hatches
> 
> it did not do that for me on Linux with 0.2.1

Is litert 3.0 published? https://crates.io/crates/litert/versions

```
error: failed to select a version for the requirement `litert = "^0.3"`
candidate versions found which didn't match: 0.2.1, 0.2.0, 0.1.0
location searched: crates.io index
```

build.rs
shubhamg13: I just modified assuming it it make clear as am creating a helper function that mentions `coreml_protos`. 

Should I move it back here?

theHamsta: The comment explains why ONNX is not compiled (it is already in `webnn-onnx-utils`).

So why ONNX not is the important part of the comment. We could make this clearer.

Makefile
theHamsta: no. it is WIP. I just wanted to make clear that I read the git docs and was not sure whether this is also true for the released 0.2.1 version.

build.rs
shubhamg13: We can get rid of flatc if we upload the generated file. WDYS?

build.rs
shubhamg13: Adding it back.

src/backend_selection.rs
shubhamg13: Sure, Updated.

.github/workflows/ci.yml
shubhamg13: Sure, I will work on it.

src/backends/litert.rs
shubhamg13: Added to `LiteRtContext` constructor. Let me know if it is right way.

build.rs
theHamsta: I think
```
run_flatc("protos/tflite/schema.fbs", &out_dir).expect("Running flatc failed. Make sure flatc is installed an in PATH");
println!("cargo:rerun-if-changed=protos/tflite/schema.fbs");
```
would be fine. Then, it fails with the explanation that `flatc` failed and not on the inclusion of the missing generated file

.github/workflows/ci.yml
theHamsta: @shubhamg13 I think we can also just enable https://github.com/rustnn/rustnn/pull/151 for `litert`

build.rs
shubhamg13: Sure integrated and resolving.

.github/workflows/ci.yml
shubhamg13: For some reason current Conformance Gate started failing. Then I guess I should wait for #151 to be merged first?

shubhamg13: Add `flatc` to `RustNNPT Conformance Gate`

.github/workflows/ci.yml
shubhamg13, line 243: I am curious about this: Since we are only running `RustNNPT Conformance Gate` on onnx, so why do we need protobuf? 

Can we just add a `cfg` flag in `build.rs`?

theHamsta: add the moment `protobuf` is a hard requirement of webnn-graphs, so to build RustNN itself. We should make onnx/protobuf optional in the long-term.

protoc is used in webnn-graph to generate the onnx schema. similar like you use `flatc` for litert

shubhamg13:
Is there anything that needs attention in this PR or it is good to go?

theHamsta:
> Is there anything that needs attention in this PR or it is good to go?

I get still SEGFAULTs when executing the tests. What should I do to run the tests successfully on Linux?

It's a SEGFAULT on mulit-threaded test runs.

theHamsta:
Also tests like

```
mlgraphbuilder::test::add_mat_plus_scalar
mlgraphbuilder::test::unused_incompatible_inputs
```
are failing. You could disable for litert for now.

build.rs
```diff
@@ -20,17 +22,39 @@ fn recurse(dir: &Path, files: &mut Vec<String>) {
     }
 }
 
-fn main() -> Result<(), Box<dyn std::error::Error>> {
+fn run_flatc(schema: &str, out_dir: &str) -> Result<(), String> {
```
theHamsta, line 28: I think `flatc` should only be required with the litert feature turned on (we should do the same for protoc for coreml/onnx)

shubhamg13:
> > Is there anything that needs attention in this PR or it is good to go?
> 
> I get still SEGFAULTs when executing the tests. What should I do to run the tests successfully on Linux?
> 
> It's a SEGFAULT on mulit-threaded test runs.

IMO it is related to `LiteRT`. Should we file ann issue for same? What do you say?

theHamsta:
> > > Is there anything that needs attention in this PR or it is good to go?
> > 
> > 
> > I get still SEGFAULTs when executing the tests. What should I do to run the tests successfully on Linux?
> > It's a SEGFAULT on mulit-threaded test runs.
> 
> IMO it is related to `LiteRT`. Should we file ann issue for same? What do you say?

Does it crash for you? Which platform are you testing? Would it be possible to work around the issue?

I can avoid the crash when I just use the CPU backend (disable webgpu via Vulkan). I don't know whether this is just not the case because the CPU backend is faster. The crash with Vulkan occurs both when restricting Vulkan drivers to llavapipe or when using the Nvidia drivers.

I also still get warnings on the file generated by flatc.

theHamsta:
> > > > Is there anything that needs attention in this PR or it is good to go?
> > > 
> > > 
> > > I get still SEGFAULTs when executing the tests. What should I do to run the tests successfully on Linux?
> > > It's a SEGFAULT on mulit-threaded test runs.
> > 
> > 
> > IMO it is related to `LiteRT`. Should we file ann issue for same? What do you say?
> 
> Does it crash for you? Which platform are you testing? Would it be possible to work around the issue?
> 
> I can avoid the crash when I just use the CPU backend (disable webgpu via Vulkan). I don't know whether this is just not the case because the CPU backend is faster. The crash with Vulkan occurs both when restricting Vulkan drivers to llavapipe or when using the Nvidia drivers.
> 
> I also still get warnings on the file generated by flatc.

Reporting to litert-rs would be good. I can still repro with their git code.

shubhamg13:
> > > > > Is there anything that needs attention in this PR or it is good to go?
> 
> > > > 
> 
> > > > 
> 
> > > > I get still SEGFAULTs when executing the tests. What should I do to run the tests successfully on Linux?
> 
> > > > It's a SEGFAULT on mulit-threaded test runs.
> 
> > > 
> 
> > > 
> 
> > > IMO it is related to `LiteRT`. Should we file ann issue for same? What do you say?
> 
> > 
> 
> > Does it crash for you? Which platform are you testing? Would it be possible to work around the issue?
> 
> > 
> 
> > I can avoid the crash when I just use the CPU backend (disable webgpu via Vulkan). I don't know whether this is just not the case because the CPU backend is faster. The crash with Vulkan occurs both when restricting Vulkan drivers to llavapipe or when using the Nvidia drivers.
> 
> > 
> 
> > I also still get warnings on the file generated by flatc.
> 
> 
> 
> Reporting to litert-rs would be good. I can still repro with their git code.

Yeah, it did crash on my Linux System. But works fine with single thread except those two tests, as you mentioned.

But I didn't test with only CPU accelerator as you highlighted.

Also, I tried to use Mutex Lock for creating env for litert, but didn't succeed. So, I think it's a bug in litert-sys

I will file an issue by collecting some more detail about it soon.

theHamsta:
> > > > > > Is there anything that needs attention in this PR or it is good to go?
> > 
> > 
> > > > >
> > 
> > 
> > > > >
> > 
> > 
> > > > > I get still SEGFAULTs when executing the tests. What should I do to run the tests successfully on Linux?
> > 
> > 
> > > > > It's a SEGFAULT on mulit-threaded test runs.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > IMO it is related to `LiteRT`. Should we file ann issue for same? What do you say?
> > 
> > 
> > > 
> > 
> > 
> > > Does it crash for you? Which platform are you testing? Would it be possible to work around the issue?
> > 
> > 
> > > 
> > 
> > 
> > > I can avoid the crash when I just use the CPU backend (disable webgpu via Vulkan). I don't know whether this is just not the case because the CPU backend is faster. The crash with Vulkan occurs both when restricting Vulkan drivers to llavapipe or when using the Nvidia drivers.
> > 
> > 
> > > 
> > 
> > 
> > > I also still get warnings on the file generated by flatc.
> > 
> > 
> > Reporting to litert-rs would be good. I can still repro with their git code.
> 
> Yeah, it did crash on my Linux System. But works fine with single thread except those two tests, as you mentioned.
> 
> But I didn't test with only CPU accelerator as you highlighted.
> 
> Also, I tried to use Mutex Lock for creating env for litert, but didn't succeed. So, I think it's a bug in litert-sys
> 
> I will file an issue by collecting some more detail about it soon.

can you avoid problems by skipping the right tests for the litert backend? I think the problem is one of the tests panicking while the other is running.

@mtavenrath has pushed a more comprehensive test suite to main that should iterate overall backends.

.github/workflows/ci.yml
shubhamg13: Enable WPT Conformance Test with passing rate of about 19% as of now. (Will work on more pending operators in follow-up)

.github/workflows/ci.yml
shubhamg13: Gotcha. Thanks for explaining.

shubhamg13:
> > > > > > > Is there anything that needs attention in this PR or it is good to go?
> > > 
> > > 
> > > > > >
> > > 
> > > 
> > > > > >
> > > 
> > > 
> > > > > > I get still SEGFAULTs when executing the tests. What should I do to run the tests successfully on Linux?
> > > 
> > > 
> > > > > > It's a SEGFAULT on mulit-threaded test runs.
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > >
> > > 
> > > 
> > > > > IMO it is related to `LiteRT`. Should we file ann issue for same? What do you say?
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > Does it crash for you? Which platform are you testing? Would it be possible to work around the issue?
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > I can avoid the crash when I just use the CPU backend (disable webgpu via Vulkan). I don't know whether this is just not the case because the CPU backend is faster. The crash with Vulkan occurs both when restricting Vulkan drivers to llavapipe or when using the Nvidia drivers.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > I also still get warnings on the file generated by flatc.
> > > 
> > > 
> > > Reporting to litert-rs would be good. I can still repro with their git code.
> > 
> > 
> > Yeah, it did crash on my Linux System. But works fine with single thread except those two tests, as you mentioned.
> > But I didn't test with only CPU accelerator as you highlighted.
> > Also, I tried to use Mutex Lock for creating env for litert, but didn't succeed. So, I think it's a bug in litert-sys
> > I will file an issue by collecting some more detail about it soon.
> 
> can you avoid problems by skipping the right tests for the litert backend? I think the problem is one of the tests panicking while the other is running.
> 
> @mtavenrath has pushed a more comprehensive test suite to main that should iterate overall backends.

Sure added list for unsupported ops : https://github.com/rustnn/rustnn/pull/149/changes/8f0ce94f963201e522411d37b2aa95b492fdd5b0

src/backends/litert.rs
```diff
@@ -127,44 +388,70 @@ impl<'context> MLBackendContext<'context> for LiteRtContext {
...
-        if array.len() > host.len() {
+        let logical = tensor.descriptor().rustnn_required_bytes();
+        if array.len() < logical {
             return Err(Error::TensorWriteError {
                 source: format!(
-                    "write exceeds tensor storage: {} bytes > {}",
+                    "write too small for tensor: {} bytes < {} logical bytes",
                     array.len(),
-                    host.len()
+                    logical,
```
theHamsta, line 402: shouldn't we have errors for any mismatches (too small and too large)?

but should already be catches by the API shared cross-backends.

theHamsta (dismissed):
LGTM.

We can add the litert tests to CI in a follow-up (e.g. after #171) and try to integrate litert logging with the log create.

shubhamg13:
> LGTM.
> 
> We can add the litert tests to CI in a follow-up (e.g. after #171) and try to integrate litert logging with the log create.

Sure, Reverted the CI Conformance task and Snapshosts. Please help to merge.

theHamsta:
> > LGTM.
> > We can add the litert tests to CI in a follow-up (e.g. after #171) and try to integrate litert logging with the log create.
> 
> Sure, Reverted the CI Conformance task and Snapshosts. Please help to merge.

No, was all good.

For me it was ready to merge with the CI included. I was assuming the WPT tests needed some change to not fail on GH actions, but all fine if everything is green.

theHamsta:
> > LGTM.
> > We can add the litert tests to CI in a follow-up (e.g. after #171) and try to integrate litert logging with the log create.
> 
> Sure, Reverted the CI Conformance task and Snapshosts. Please help to merge.

The snapshot also matches what I would get locally. So I would like to include them. I reverted your revert. For me good to merge.

theHamsta (approved):

shubhamg13:
> > > LGTM.
> > > We can add the litert tests to CI in a follow-up (e.g. after #171) and try to integrate litert logging with the log create.
> > 
> > 
> > Sure, Reverted the CI Conformance task and Snapshosts. Please help to merge.
> 
> The snapshot also matches what I would get locally. So I would like to include them. I reverted your revert. For me good to merge.

Sure, Please help me to merge

theHamsta:
Thank you for your contribution! Please excuse any delays on our side!

PR #152, opened 2026-06-21
feat(litert): Add placeholder for litert as backend

shubhamg13:
## Summary

This just adds a placeholder for LiteRt as backend option. 
Complete implementation will come in follow-up PR.

## Validation

- [x] `make test`

## Documentation

### DIFF
--- a/src/backends/litert.rs
+++ b/src/backends/litert.rs
@@ -1,0 +1,222 @@
+use std::collections::HashMap;
+use std::fmt;
+
+use crate::GraphInfo;
+use crate::backend_selection::DeviceType;
+use crate::error::{Error, Result};
+use crate::mlcontext::{
+    ListDevices, MLBackendBuilder, MLBackendContext, MLGraph, MLTensor, MLTensorDescriptor,
+};
+
+pub(crate) struct LiteRtTensor {
+    memory: Vec<u8>,
+}
+
+pub(crate) struct LiteRtGraph;
+
+impl fmt::Debug for LiteRtGraph {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("LiteRtGraph").finish()
+    }
+}
+
+pub(crate) struct LiteRtContext {
+    tensors: Vec<LiteRtTensor>,
+    device_type: DeviceType,
+}
+
+impl fmt::Debug for LiteRtContext {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("LiteRtContext")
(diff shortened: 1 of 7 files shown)


shubhamg13:
cc: @xiaochengh

shubhamg13:
This change is ready for review. Please take a look and let me know your thoughts.

src/backends/litert.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    fn rustnn_set_tensor_capacity(
+        &mut self,
+        _tensor: &mut MLTensor,
+        _max_shape: &[u64],
+    ) -> Result<()> {
+        Ok(())
+    }
+
+    fn create_constant_tensor(
```
theHamsta, line 101: all backends have implemented this method in the same way. We should hoist this as a default impl to the trait as a follow-up. (also with MLOperandDescriptor instead of MLTensorDescriptor). I have this already prepared. Just wanted to give a FYI.

At the moment, constant tensors are not handled correctly by RustNN. RustNN does not enforce that constant tensors are never readable and writable.

theHamsta (dismissed):

src/backends/litert.rs
shubhamg13: Yeah, I also observed this that we can move bunch of code to trait. Thanks for explaining.

theHamsta: Also if you feel that something in the traits for the backends could be done better to accommodate for capabilities of LiteRT or other backends: feel free to suggest improvements!

One thing that I'm planning is to give the backends a way that they can accept created constants from MLGraphBuilder::constant or MLContext::create_constant_tensor directly without the need to store them in GraphInfo if they don't want to.

shubhamg13: Is it good to go for now...

src/backends/litert.rs
```diff
@@ -0,0 +1,222 @@
...
+        self.tensors.push(LiteRtTensor { memory });
+        Ok(MLTensor {
+            id: self.tensors.len() - 1,
+            constant: false,
+            descriptor: descriptor.clone(),
+        })
+    }
+
+    fn rustnn_resize_tensor(&mut self, _tensor: &mut MLTensor, _new_shape: &[u64]) -> Result<()> {
+        Ok(())
```
theHamsta: could you `todo!()`/`panic!()` or error out for unsupported methods?

shubhamg13: Added `todo!` for unimplemented methods

theHamsta: thank you!

theHamsta:
Thank you for contributing the LiteRT backend! This will enable much more devices than we could support before

PR #162, opened 2026-06-26
feat: add builder API sample + trtx improvements

theHamsta:
## Summary

- Add builder API sample (ye old style transfer)
- register host pointers in trtx backend
- capture CUDA graphs in trtx backend

## Validation

- [x] `make test`

## Documentation

### DIFF
--- a/examples/fast_style_transfer_builder_api.rs
+++ b/examples/fast_style_transfer_builder_api.rs
@@ -1,0 +1,983 @@
+// Port of https://d3i5xkfad89fac.cloudfront.net/test-data/models/fast_style_transfer_nchw/weights
+// (Apache 2.0)
+use std::{collections::HashMap, fmt, path::PathBuf, time::Instant};
+
+use anyhow::{Context, Result, anyhow, bail, ensure};
+use clap::{Parser, ValueEnum};
+use futures_util::future::join_all;
+use rustnn::mlcontext::{
+    MLContext, MLContextOptions, MLGraphBuilder, MLOperand, MLOperandDescriptor, MLPowerPreference,
+    MLTensorDescriptor,
+};
+use rustnn::operator_enums::MLOperandDataType;
+use rustnn::operator_options::{
+    MLConv2dOptions, MLConvTranspose2dOptions, MLPadOptions, MLReduceOptions,
+};
+use tokio::sync::Mutex;
+
+const DEFAULT_WEIGHTS_BASE_URL: &str = "https://raw.githubusercontent.com/webmachinelearning/test-data/0495fc5b5e4ccf77f745b747aa43e12a71a30cff/models/fast_style_transfer_nchw/weights/";
+
+const BATCH: usize = 1;
+const CHANNELS: usize = 3;
+const HEIGHT: usize = 540;
+const WIDTH: usize = 540;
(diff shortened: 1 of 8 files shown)


src/graph.rs
```diff
@@ -109,14 +109,14 @@ impl DataType {
...
-        Some((total_bits + 7) / 8)
+        Some(total_bits.div_ceil(8))
     }
 
     /// Byte size of one element when the type is whole-byte aligned (`bits % 8 == 0`).
     pub fn bytes_per_element(self) -> usize {
         let bits = self.bits_per_element();
         debug_assert!(
-            bits % 8 == 0,
+            bits.is_multiple_of(8),
```
theHamsta, line 119: clippy

src/graph.rs
```diff
@@ -109,14 +109,14 @@ impl DataType {
     /// Host storage bytes for `elements` values (always `ceil(elements * bits / 8)`).
     pub fn storage_byte_length(self, elements: usize) -> Option<usize> {
         let total_bits = elements.checked_mul(self.bits_per_element())?;
-        Some((total_bits + 7) / 8)
+        Some(total_bits.div_ceil(8))
```
theHamsta, line 112: this is finally in stable 👀

src/backends/trtx.rs
```diff
@@ -139,6 +141,57 @@ pub(crate) struct TrtxGraph<'context> {
...
+    cuda_graphs: HashMap<Vec<(bool, String, usize)>, CapturedCudaGraph>,
+}
+
+struct CapturedCudaGraph {
+    graph: sys::CUgraph,
+    executable: sys::CUgraphExec,
+    stream: Arc<CudaStream>,
+}
+
+impl CapturedCudaGraph {
```
theHamsta, line 153: Graph capture and pinned pointers needed for 100% GPU usage.

long-term, `write_tensor`/`read_tensor` should also be fixed to have async semantics (return quickly, signal when done, e.g. via `async` or by returning a sync token to wait for). Direct access to tensor accelerator memory also needed.

examples/fast_style_transfer_builder_api.rs
```diff
@@ -0,0 +1,1054 @@
...
+    fn get(&self, name: &str) -> Result<MLOperand> {
+        self.weights
+            .iter()
+            .find(|weight| weight.name == name)
+            .map(|weight| weight.operand)
+            .with_context(|| format!("missing loaded weight `{name}`"))
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
mtavenrath, line 211: A numpy file parser might be usable for other samples as well. Should this be in a separate file?

src/backends/trtx.rs
mtavenrath: I assume capturing the CUDA graph means that dynamic shapes are unsupported in this mode?

mtavenrath (approved):

theHamsta:
https://github.com/rustnn/rustnn/pull/162#discussion_r3497286537

We would need to add the shape to the cache key. TRTX does not considering dynamic shapes at the moment. It would need to set input shapes and also the converter would need to do dynamic shapes

examples/fast_style_transfer_builder_api.rs
theHamsta: I also thought that but it would need to refactor the sample to have their own cargo file or have a dedicated numpy package.

PR #164, opened 2026-06-30
feat!: return sync handle for read_tensor/write_tensor

theHamsta:
## Summary

We have three choices to implement `async` WebNN APIs:
1) just make them blocking, when they return they are finished
2) return a sync handle that can be waited on blocking or polled if finished
3) return a future to implement a proper async API (potentially by turning 2) into futures)
4) push blocking work into background thread, get notified when finished. use that to implement async

Things to consider:
- `async` traits don't play well with dynamic dispatch, backends might need to return an object from a non-async trait that the RustNN API turns into a future
- futures must be able to wake/notify when they are ready, futures can be canceled
- we likely want the `MLGraphBuilder::build` to be cancelable because it potentially takes a long time
- we likely want `read_tensor`/`write_tensor` to not be 1) because otherwise the GPU has to synchronize on each of the calls and for example for `write_tensor` the host does not know when it is safe to write 
- we can't really borrow in `read_tensor`/`write_tensor` as host could drop the arrays in the meantime. Would need to hold on the reference for a while via Arc or a different API for `read_tensor`/`write_tensor` that allows direct access to the internal handle of the tensor  

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/backends/coreml.rs
+++ b/src/backends/coreml.rs
@@ -21,6 +21,7 @@ use crate::executors::coreml::{
 };
 use crate::mlcontext::{
     MLBackendBuilder, MLBackendContext, MLBackendGraph, MLGraph, MLTensor, MLTensorDescriptor,
+    SyncHandle,
 };
 
 /// Number of bytes required to store a tensor described by `descriptor`.
@@ -135,7 +136,11 @@ impl<'context> MLBackendContext<'context> for CoremlContext {
         Ok(tensor)
     }
 
-    fn read_tensor(&mut self, tensor: &MLTensor, array: &mut [u8]) -> crate::error::Result<()> {
+    fn read_tensor(
+        &mut self,
+        tensor: &MLTensor,
+        array: &mut [u8],
+    ) -> crate::error::Result<SyncHandle> {
         let host = &self.tensors[tensor.id].memory;
         let logical = tensor_byte_len(tensor.descriptor());
         if array.len() < logical {
@@ -154,10 +159,14 @@ impl<'context> MLBackendContext<'context> for CoremlContext {
             tensor: tensor.clone(),
         })?;
         array[..logical].copy_from_slice(slice);
-        Ok(())
+        Ok(SyncHandle::Ready)
     }
 
-    fn write_tensor(&mut self, tensor: &MLTensor, array: &[u8]) -> crate::error::Result<()> {
+    fn write_tensor(
+        &mut self,
+        tensor: &MLTensor,
+        array: &[u8],
+    ) -> crate::error::Result<SyncHandle> {
         let host = &mut self.tensors[tensor.id].memory;
         if array.len() > host.len() {
             return Err(Error::TensorWriteError {
@@ -172,7 +181,7 @@ impl<'context> MLBackendContext<'context> for CoremlContext {
         }
         let n = array.len();
         host[..n].copy_from_slice(array);
-        Ok(())
+        Ok(SyncHandle::Ready)
     }
 
     fn dispatch(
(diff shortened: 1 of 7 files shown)


src/backends/coreml.rs
```diff
@@ -135,7 +136,11 @@ impl<'context> MLBackendContext<'context> for CoremlContext {
         Ok(tensor)
     }
 
-    fn read_tensor(&mut self, tensor: &MLTensor, array: &mut [u8]) -> crate::error::Result<()> {
+    fn read_tensor(
+        &mut self,
+        tensor: &MLTensor,
+        array: &mut [u8],
+    ) -> crate::error::Result<SyncHandle> {
```
theHamsta, line 143: it is both for `read_tensor` and `write_tensor` to return without the transfer being finished, because the borrow of the host array is limited to the function call, but caller could drop the array before transaction finished.

Would need to be Arc or permanent borrow

theHamsta: permanent borrow for the lifetime of the tensor would also allow for unified memory, but `read_tensor` with its `&mut [u8]` can only have an exclusive reference, as the content is modified. So, read_tensor would still need to sync.

matthargett:
From the watchOS side, I think the important abstraction here is ownership and hazards, not just “copy completion.” watchOS 11 adds Core ML [MLTensor(bytesNoCopy:...)](https://developer.apple.com/documentation/coreml/mltensor/init%28bytesnocopy%3Ashape%3Ascalartype%3Adeallocator%3A%29); the MLModel path also has [MLMultiArray(dataPointer:...)](https://developer.apple.com/documentation/coreml/mlmultiarray/init%28datapointer%3Ashape%3Adatatype%3Astrides%3Adeallocator%3A%29) and [outputBackings](https://developer.apple.com/documentation/coreml/mlpredictionoptions/outputbackings). Unified memory can remove a physical copy, but CoreML may still retain or use that backing asynchronously, and CPU/GPU/ANE visibility still needs an ordering point to rally around.

it seems like a handle returned after an &[u8]/&mut [u8] borrow ends is only safe *if* it retains the allocation (an Arc/owned buffer), or the borrow is scoped through completion. does that track from your understanding also? 

I’d also separate input-consumed from output-ready; their reuse rules differ. Apple’s watchOS 11 [MLState](https://developer.apple.com/documentation/coreml/mlstate) contract is a relevant (to me) precedent: no buffer access while prediction is in flight, and predictions sharing state must serialize. does that make target workloads too slow on bigger iron?

theHamsta:
> From the watchOS side, I think the important abstraction here is ownership and hazards, not just “copy completion.” watchOS 11 adds Core ML [MLTensor(bytesNoCopy:...)](https://developer.apple.com/documentation/coreml/mltensor/init%28bytesnocopy%3Ashape%3Ascalartype%3Adeallocator%3A%29); the MLModel path also has [MLMultiArray(dataPointer:...)](https://developer.apple.com/documentation/coreml/mlmultiarray/init%28datapointer%3Ashape%3Adatatype%3Astrides%3Adeallocator%3A%29) and [outputBackings](https://developer.apple.com/documentation/coreml/mlpredictionoptions/outputbackings). Unified memory can remove a physical copy, but CoreML may still retain or use that backing asynchronously, and CPU/GPU/ANE visibility still needs an ordering point to rally around.
> 
> it seems like a handle returned after an &[u8]/&mut [u8] borrow ends is only safe _if_ it retains the allocation (an Arc/owned buffer), or the borrow is scoped through completion. does that track from your understanding also?
> 
> I’d also separate input-consumed from output-ready; their reuse rules differ. Apple’s watchOS 11 [MLState](https://developer.apple.com/documentation/coreml/mlstate) contract is a relevant (to me) precedent: no buffer access while prediction is in flight, and predictions sharing state must serialize. does that make target workloads too slow on bigger iron?

Yes, I also realized that we need some wrapper around the host buffer that would avoid hazard (a bit but not quite like in https://github.com/rustnn/rustnn/pull/174, I don't like the solution there yet). A wrapper around the host memory could then also allow unified memory by making GPU tensors just take a borrow on a host memory object and only do a copy-on-write when the host memory is about to get modified independently of the GPU tensor.

With sync handles, even for separate input consumed/output ready we can't have a safe API.

PR #151, opened 2026-06-20
Implement WPT-WebNN test harness in RustNN

mtavenrath:
## Summary

rustnnpt used the old deprecated interface to construct the graph lacking test for the actual official MLBuilder interface. Also having the test harness in a different project frequently lead to problems. Thus this change uses node.js to extract all required test information and build the graph using the MLBuilder API.

A positive side effect is that test evaluation can now run a few seconds when using the ONNX CPU backend.

## Validation

- [ ] `make test`
- [x] Relevant WPT or integration checks

## Documentation

- [x] Updated docs if behavior changed
- [ ] If backend converter/executor operator support changed, ran `make docs-backend-ops` and committed `docs/development/backend-operator-support.md`

PR #165, opened 2026-06-30
feat(trtx): use jit cache

theHamsta:
## Summary

Add JIT cache for TRTX. This helps with JIT compiling CUDA kernel. This provides smaller benefit that the caching of engines (which is still disabled at the moment). This cache is good for the specialization of an engine for a specific GPU. Building an engine is a step that can be done on CPU without needing a GPU or a GPU driver.

ONNX runtime EP has one cache per ONNX. We try with a global cache. Just to collect different experiences.

pro global cache: cache can be shared between kernels
con global cache: cache can grow too big which can have negative perf implication. TRT docs recommend to not drop cache when growing to big

Needs to be validated if useful with #151 

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/backends/trtx.rs
+++ b/src/backends/trtx.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
-use std::ffi::c_void;
+use std::ffi::{CStr, c_void};
 use std::mem::MaybeUninit;
 use std::ops::Deref;
 use std::rc::Rc;
@@ -223,6 +223,7 @@ impl TrtxTensor {
 
 pub(crate) struct TrtxContext<'context> {
     cuda_ctx: Arc<CudaContext>,
+    pci_device_id: String,
     tensors: Vec<TrtxTensor>,
     events: Vec<CudaEvent>,
     attempted_host_registrations: HashSet<usize>,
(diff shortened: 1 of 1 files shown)


src/backends/trtx.rs
```diff
@@ -323,6 +389,26 @@ static TRTX_SUFFIX: LazyLock<String> = LazyLock::new(|| {
...
 
+fn jit_cache_disk_key(pci_device_id: &str) -> String {
+    let filesystem_safe_device_id = pci_device_id.replace([':', '.'], "_");
+    format!("{filesystem_safe_device_id}_{}.cache", *TRTX_SUFFIX)
+}
+
+// RuntimeCache owns its TensorRT IRuntimeCache and does not borrow the engine. The lifetime in
+// trtx::RuntimeCache is represented only by PhantomData, so erase it while the cache is held in
+// the process-global map and restore the current engine lifetime when cloning it out.
+fn into_global_runtime_cache<'engine>(
```
theHamsta, line 400: not nice. but will work for experiments.

theHamsta: trtx-rs is responsible to provide ergonomic API in the end

theHamsta:
Closed for #193

PR #169, opened 2026-07-01
TRT-RTX test failure sweep

mtavenrath:
## Summary

AI went through all the test failures and fixed those. Some of the fixes are most likely not ideal, yet the current goal is to achieve conformance before attempting to achieve performance.

## Validation

- [ ] `make test`

## Documentation


src/converters/trtx.rs
```diff
@@ -268,38 +565,237 @@ impl TrtxConverter {
             })
     }
 
-    /// Cast INT32 tensor to Float32
-    fn cast_int32_to_float32<'a>(
+    /// TensorRT `IPaddingLayer` / rank-shuffle pad path dtypes (not UINT8/INT32/INT64).
+    fn trtx_ipadding_layer_supports_dtype(dtype: TrtDataType) -> bool {
```
theHamsta, line 569: the official way to do more general padding is to do ISliceLayer with SampleMode::kFILL (you can specify a fill value). You do slice out-of-bounds with negative indices and indices larger than tensor dimensions

not concat.

theHamsta: could be merged with this as a TODO.

src/converters/trtx.rs
```diff
@@ -2514,6 +3060,250 @@ impl TrtxConverter {
...
+                    format: "trtx".to_string(),
+                    reason: format!(
+                        "{label}: no TensorRT or WebNN static dimensions for operand {operand_id}"
+                    ),
+                }),
+        }
+    }
+
+    /// True when a graph constant integer tensor is all zeros (common Q/DQ zero_point).
+    fn trtx_constant_integer_is_all_zero(
```
theHamsta, line 3091: I have a change where I take away all weights from trtx to refit them later.

Would it be possible to classify constant reads in this PR to "I really need the value"/impossible to refit and let them go through a dedicated function call?

Then, it would be easier for me to take away access where it is not needed.

theHamsta: even better would be if the detection on whether a weight needs to be available or can be uploaded to GPU can be done during network build, but I assume we try to make almost all not needed for building.

src/graph.rs
```diff
@@ -123,15 +123,16 @@ impl DataType {
     }
 }
 
-/// Packed int4/uint4 layout: even logical indices in the high nibble, odd in the low.
+/// Packed int4/uint4 layout: even logical indices in the low nibble, odd in the high (ONNX/WebNN
+/// convention, i.e. element `2*i` occupies the least-significant 4 bits of byte `i`).
 pub fn unpack_int4(data: &[u8], element_count: usize) -> Vec<i32> {
```
theHamsta, line 128: packing and unpacking is still used in the trtx converter and here

theHamsta (dismissed):

src/converters/trtx.rs
mtavenrath: files #178

PR #171, opened 2026-07-01
Add macOS CI

mklimenko-nv:
rustnn targets macOS and CoreML, but has no CI to catch build failures on that platform.
This PR adds macOS coverage to CI. As of opening, it is expected to fail, since the CoreML backend has never been exercised in CI before.

Besides enabling CI, the following fixes were made:
- Fix default/no-feature build, where `CoremlContext` and `LiteRtContext` both aliased `DisabledContext`, resulting in a duplicate `new_from_device_type`
- Fix `-D warnings` on macOS to avoid `cargo-clippy` issues
- Fix SIGSEGV caused by invalid dtype codes in CoreML path
- Added `.mlpackage` manifest, necessary for the weight-bearing models
- Added an exception firewall to catch CoreML-issues Objective-C exceptions
- Isolate WPT runs to gracefully handle hard faults (like null-deref in Apple's libBNNS on transpose-of-constant).

Known limitations:
Current pass rate is about 56%, with issues like unimplemented converter, `int4`/`uint4`, incorrect rank for quantize ops, etc. Will be fixed in a separate PR.


mklimenko-nv:
Even after resolving the temporary paths, there is a difference between my local M5 and a GitHub-based M1:
https://github.com/rustnn/rustnn/actions/runs/28580149358/job/84738188993?pr=171#step:10:862

src/executors/coreml.rs
```diff
@@ -26,6 +25,57 @@ unsafe extern "C" {}
...
+        err_len: usize,
+    ) -> i32;
+}
+
+/// Make a CoreML error message deterministic so it doesn't churn snapshots:
+/// - Replace the temp-model path CoreML embeds in some errors (e.g. the `-14`
+///   execution-plan failure); it carries a timestamp/UUID.
+/// - Sort the allowed-type set CoreML lists in `Expected { ... }` clauses, which
+///   it otherwise enumerates in a nondeterministic (hash) order.
+fn sanitize_coreml_error(msg: &str) -> String {
```
theHamsta: isn't this something we should just do in run_wpt_test?

I assume the temp files are not removed when coreml is crashing, so they could be useful if you wanted to manually inspect them.

but probably `<coreml-temp-model>` is fine also

mklimenko-nv: That's fair, let's keep it in one place, fixed in 02009e3

theHamsta:
@mtavenrath would need to go to repo settings to change the required CI tests since the generic "Rust Tests" now got renamed to "Rust Tests (ubuntu-latest)" and "Rust Test (macos-latest)"

theHamsta (dismissed):

theHamsta:
Minimal tests (e.g. build and lib tests) are currently passing. But snapshots can't be used for WPT tests as failing/passing tests for coreml seem to depend on target device, so we need to work with a skip or allowed-to-fail list to make wpt pass

mklimenko-nv:
Closed in favor of #172

PR #172, opened 2026-07-03
Macos ci fixup

theHamsta:
## Summary

Some additional changes on top of #171 to make CI green

## Validation

- [ ] `make test`

## Documentation


theHamsta:
coreml builds also on linux when using `https://github.com/gnustep/libobjc2` (otherwise you get a missing `objc_autoreleasePoolPop` linker error when using gcc libobj). When trying to force coreml on Linux during runtime, you get a "Class with name NSString XXX" not found error. Since the `have_coreml` stays false on Linux, coreml will never be selected even if someone enables the `coreml-feature`.

It's useful to build this on Linux, because you will see build errors for all backend and can do refactorings with all backends enabled.

theHamsta:
Also this (https://crates.io/crates/objc2 https://github.com/madsmtm/objc2) is apparently the modern alternative to the `objc` crate (last release 6 years ago) https://github.com/SSheldon/rust-objc

theHamsta:
The coreml backend has a concurrency problem. Presumably related to the temporary file names conflicting. We should revise that in a follow-up.

The snapshots also match on M5 (were originally sourced from a M1).

There seems to be still one test that can SEGFAULT

theHamsta:
#173 the issue with the failing tests in multi-threaded mode comes from the tempfiles whose names are conflicting. Can be resolved by doing everything in-memory.

theHamsta:
image

Test name changed to add ` (ubuntu-latest)` suffix therefore required tests in repo settings need to be changed

theHamsta:
Closing in favor of #173 (which fixes the SEGFAULT that can still occur with this pipeline

PR #174, opened 2026-07-06
Into the future: async

theHamsta:
## Summary

Try to make RustNN async, see how it goes

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/backends/trtx.rs
+++ b/src/backends/trtx.rs
@@ -28,7 +28,7 @@ use crate::converters::TrtxConverter;
 use crate::error::Error;
 
 use crate::error::GraphBuilderError;
-use crate::mlcontext::MLTensor;
+use crate::mlcontext::{HostBuffer, MLTensor, StreamCallbackState, TensorSyncHandle};
 use crate::mlcontext::{ListDevices, MLOperand};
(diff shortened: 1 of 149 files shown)


src/backends/trtx.rs
```diff
@@ -240,13 +365,42 @@ impl TrtxTensor {
...
 }
 
+struct ReadTensorCallback {
+    state: Arc<StreamCallbackState>,
+    buffer: HostBuffer,
+}
+
+struct WriteTensorCallback {
+    state: Arc<StreamCallbackState>,
+    buffer: HostBuffer,
```
theHamsta, line 375: this seems to be slower to sync as when waiting on it blocking.

The futex waiting on the cuda transfer to be ready is slower than spinning on the result (50us latency on futex on linux)

theHamsta: with `cuLaunchHostFunc_v2`'s `CU_HOST_TASK_SPINWAIT` additional latency goes down to ~10us

theHamsta:
The question with async is also whether we would need a tokio feature that would use tokio mutex and tokio oneshot channels etc. Or whether the accelerator async is the same kind of async as graph compile async or IO async

PR #181, opened 2026-07-13
feat: introduce RustNNOptions

theHamsta:
## Summary

Allows to have both global and backend-specific options.

At the moment strongly-typed with all public members. This will make adding/removing options a breaking change which we might want in the bring-up phase of RustNN.

Long-term stability could come from either:
- builder methods for options
- untyped key-value options
- strongly typed key-value options

EDIT: `#[non_exhaustive]` would also work to make adding new options or backend non-breaking

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/src/backends/trtx.rs
+++ b/src/backends/trtx.rs
@@ -14,7 +14,7 @@ use log::debug;
 use log::info;
 use log::trace;
 use log::warn;
-use std::sync::{Arc, Mutex, OnceLock};
+use std::sync::{Arc, Mutex};
 use trtx::CudaEngine;
 use trtx::ExecutionContext;
 use trtx::Refitter;
@@ -29,9 +29,11 @@ use crate::error::Error;
 
 use crate::error::GraphBuilderError;
 use crate::mlcontext::MLTensor;
+use crate::mlcontext::RustNNOptions;
 use crate::mlcontext::{ListDevices, MLOperand};
 use crate::mlcontext::{MLBackendBuilder, MLGraph};
 use crate::mlcontext::{MLBackendContext, MLBackendGraph};
+use crate::mlcontextoptions::TrtxOptions;
 
 const TRTX_JSON_DUMP_PATH_ENV_VAR: &str = "TRTX_JSON_DUMP_PATH";
 
@@ -196,18 +198,6 @@ impl Drop for CapturedCudaGraph {
     }
 }
 
-/// CUDA graph capture/replay around `enqueue_v3`. Disabled by default.
-/// Set `RUSTNN_TRTX_CUDA_GRAPHS=1` or `true` to enable.
-static TRTX_CUDA_GRAPHS_ENABLED: OnceLock<bool> = OnceLock::new();
-
-fn trtx_cuda_graphs_enabled() -> bool {
-    *TRTX_CUDA_GRAPHS_ENABLED.get_or_init(|| {
-        std::env::var("RUSTNN_TRTX_CUDA_GRAPHS")
-            .map(|v| v == "1" || v.eq_ignore_ascii_case("true"))
-            .unwrap_or(false)
-    })
-}
-
 fn enqueue_inference_v3(
     exec: &mut ExecutionContext<'_>,
     inference_stream: &CudaStream,
@@ -249,6 +239,7 @@ pub(crate) struct TrtxContext<'context> {
     runtime: Arc<Mutex<trtx::Runtime<'context>>>,
     config: Arc<Mutex<trtx::BuilderConfig<'context>>>, // needs to be destroyed before builder
     builder: Arc<Mutex<trtx::Builder<'context>>>,
+    options: TrtxOptions,
 }
 
 impl std::fmt::Debug for TrtxContext<'_> {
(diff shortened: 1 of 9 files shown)


theHamsta:
We should also make consistent in the repo whether it should be RustnnOptions, RustNnOptions, RustNNOptions, LitertOptions, LiteRtOptions, LiteRTOptions. Found `Litert` for current uses of capitalized litert in our repo, while the litert crate prefers `LiteRtLogSeverity`.

src/mlcontext.rs
```diff
@@ -335,6 +335,50 @@ pub enum MLPowerPreference {
...
+            // Lower CPU overhead using CUDA graphs
+            cuda_graphs: false,
+        }
+    }
+}
+
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub struct LitertOptions {
+    /// Logging severity from 0 (Silent) to 5 (Debug)
+    pub global_loggering_severity: u32,
```
theHamsta: limit all options to String, int, float or add new strong type for LoggingSeverity that does not allow invalid values. I would like to set LiteRT options even when litert disabled so not using LiteRT's own type.

src/mlcontext.rs
```diff
@@ -335,6 +335,67 @@ pub enum MLPowerPreference {
...
+            engine_caching: false,
+            // Lower CPU overhead using CUDA graphs
+            cuda_graphs: false,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+#[non_exhaustive]
+pub enum LiteRtLogSeverity {
```
theHamsta: now we have strongly typed type, but it has the same name as theirs, just different module scope.

we could also use their type iff their module is enabled, otherwise our.

EDIT: we have the same name as the C/C++ type or the Rust literrt_sys crate. There is no aliasing in names with the high-level `litert` crate which trims the `LiteRt` prefix

theHamsta:
@shubhamg13

I also saw that the C/C++ LiteRT or litert_sys provide a SinkLogger that does not print to stdout but that you need to regularly poll for messages. Could also be an option in future.

src/mlcontext.rs
shubhamg13: Can we get rid of `LiteRT` keyword. `LogSeverity` can be global to all the backends. It can be used if backends want to otherwise ignored. Just my thought.

theHamsta: > Can we get rid of LiteRT keyword. LogSeverity can be global to all the backends. It can be used if backends want to otherwise ignored. Just my thought.

all other backends use the log crate and don't need log verbosity. I would like to integrate litert also with that system if possible (e.g. the sinklogger).

If we can't avoid it it must be controlled somehow.

src/mlcontext.rs
```diff
@@ -335,6 +335,69 @@ pub enum MLPowerPreference {
...
+    /// Silence everything.
+    Silent,
+}
+
+#[derive(PartialEq, Eq, Clone, Debug)]
+#[non_exhaustive]
+pub struct LiteRtOptions {
+    /// Logging severity from Silent to Debug
+    pub global_loggering_severity: LiteRtLogSeverity,
+}
```
mtavenrath: Why do we have to define LiteRt in the global mlcontext?

theHamsta: because we want one way to define all the options.

Independently of whether backend enabled or not. For possible future backends that are not part of the RustNN repo we would need generic kv-options or dynamic trait objects.

theHamsta: could be a dedicated options module

src/mlcontext.rs
theHamsta: I checked again and onnx-runtime also has a log-level option https://docs.rs/ort/2.0.0-rc.12/ort/logging/enum.LogLevel.html but you can set log call backs for onnxruntime to behave like most other Rust module and log using the log crate.

We could make it global option but that would only control the internal logging state of the backend.

Maybe a better solution would be to set litert log level and onnx log level to the log level of the respective rustnn module

theHamsta: The option backend-specifc would make it clear that this is the option for the c++ backend library and not our litert backend.

The ORT library has different log levels and we could have their logging controlled `log` crate`  (which we might also achieve for litert at some point)

https://docs.rs/ort/2.0.0-rc.12/ort/logging/enum.LogLevel.html

mtavenrath:
> We should also make consistent in the repo whether it should be RustnnOptions, RustNnOptions, RustNNOptions, LitertOptions, LiteRtOptions, LiteRTOptions. Found `Litert` for current uses of capitalized litert in our repo, while the litert crate prefers `LiteRtLogSeverity`.

Ack. I'd like to see the renaming to LiteRt like it's done in the original library.

theHamsta:
@shubhamg13

We wanted to add `RuntimeLogLevel` and use it for onnxruntime and litert. But we found that both in ORT and litert the logging level is currently set globally and not per context. So we deferred logging options for now.

One would either
- need a global option
or
- have separate envs for onnx and literrt per context
- set litert logging level per env via https://github.com/theHamsta/rustnn/blob/abf14d300f035c60626498f6b8c4cf939442d5ce/target/debug/build/litert-sys-56c32a828c8fedab/out/bindings.rs#L775 (`litert::set_global_log_severity` sets globally and not per context)

shubhamg13 (approved):

shubhamg13:
> * need a global option

I am getting good feeling for this.

src/backends/trtx.rs
```diff
@@ -864,7 +856,7 @@ mod tests {
         let _ = pretty_env_logger::try_init();
         let devices = TrtxContext::list_devices();
         let mut context = if let [first, ..] = devices.as_slice() {
-            TrtxContext::new(*first.as_trtx_device().unwrap()).unwrap()
+            TrtxContext::new(*first.as_trtx_device().unwrap(), &RustNNOptions::default()).unwrap()
```
mtavenrath: could we make the options object optional to pass None instead of creating a default object?

mtavenrath (dismissed):

src/backends/trtx.rs
theHamsta: This API is not exposed by RustNN public to end users. The only usage of this function in non-test code passes the options. 

I changed it none-the-less for test code that create the context directly instead of setting a backend/backend device hint on the MLContextOptions

PR #183, opened 2026-07-14
Fix macOS WPT tests

mklimenko-nv:
## Summary

An agent went through the test failures and fixed most of those. 1053 tests are resolved vs `main`, now we're at 2433  tests passing out of 2482 (~98%). The remaining test failures are inherent due to CoreML/hardware limits:

- Integer extreme-value precision — CoreML computes int ops in fp32, so values near ±2³¹ and all int64 lose precision (abs/clamp/relu/neg int32/int64/uint32/uint64).
- Rank > 5 — abs_float16_6D, reshape 5D→6D.
- Pooling dilations (no MIL parameter) ×5; maxPool ceil all-padding-window (-inf vs 0).
- mlNumber int64/uint64 overflow/underflow + int→uint saturation semantics (9).
- reduce_log_sum_exp overflow/underflow (native lacks max-subtraction; stable form conflicts with CoreML scalar shapes).
- pad edge/reflection on >2 dims (CoreML constant-only), bigint value; gatherND rank-5; gather/scatterND OOB semantics; convTranspose padding edge.

## Validation

- [x] `make test`


src/executors/coreml.rs
```diff
@@ -576,19 +605,32 @@ unsafe fn fill_multiarray_from_bytes(
...
+            return Ok(());
+        }
+        let ptr: *mut c_void = msg_send![array, dataPointer];
+        if ptr.is_null() {
+            return Err(GraphError::CoremlRuntimeFailed {
+                reason: format!("MLMultiArray has no backing buffer for data type {dtype:?}"),
+            });
+        }
+        // Convert src bytes (dtype) → array bytes (canonical_code).
+        let converted = convert_input_bytes(src, dtype, canonical_code, count);
```
theHamsta, line 626: why not just bytemuck cast the slice? or just ignore the conversion? same also for the bytes from objc.

theHamsta: I see, sometimes conversion is done. Maybe in future the conversion could be done by coreml ops

src/converters/weight_file_builder.rs
```diff
@@ -1,252 +1,247 @@
-use std::collections::HashMap;
-
 use crate::error::GraphError;
+use std::collections::HashMap;
 
 const SENTINEL: u32 = 0xDEADBEEF;
 const ALIGNMENT: usize = 64;
+const FILE_VERSION: u32 = 2;
 
-/// Builds a CoreML weight file with proper alignment and metadata.
```
theHamsta, line 8: does this PR get rid of remaining uses of tempfiles or replace them with safe names for tempfile (the ort backeend uses the `tempfile` dependency for safe temp file names)

src/executors/coreml.rs
mklimenko-nv: The main problem is that the inputs/outputs can be passed only as `MLMultiArray`, which has a very limited subset of data types. For other WebNN data types we need to convert them on the host beforehand.

src/converters/weight_file_builder.rs
mklimenko-nv: This PR gets rid of tempfiles, it creates an in-memory asset loading, that doesn't write anything to the disk.
But there is a fallback: on any in-memory failure it calls `write_temp_model_with_weights`, which isn't properly unique. Let me fix it

mklimenko-nv: Addressed in 3fb36d2

theHamsta (approved):

PR #184, opened 2026-07-15
Add support to save a graph from the MLGraphBuilder

mtavenrath:
## Summary

For exporter and debugging reasons we want to use the MLGraphBuilder interface to validate and save the graph to avoid duplicated code logic. Another benefit is that we can validate the export and safe it a single export step.

To support those use cases the function rustnn_save_webnn has been added to the MLGraphBuilder interface.

## Validation

- [x] `make test`

## Documentation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ ort = { version = "=2.0.0-rc.12", optional = true, features = [
     "api-24",
 ] }
 half = "2.4"
+safetensors = "0.7"
 trtx = { version = "0.7.1", optional = true }
 litert = { version = "0.2.1", optional = true }
 litert-sys = { version = "0.2.1", optional = true }
@@ -87,7 +88,6 @@ seahash = "4.1.0"
 tokenizers = { version = "0.22", default-features = false, features = ["onig"] }
 tempfile = "3.8"
 half = "2.4"
-safetensors = "0.7"
 insta = { version = "1.47.2", features = ["yaml"] }
 image = { version = "0.25", default-features = false, features = ["jpeg", "png"] }
 dirs = "6.0.0"
--- a/src/webnn_json.rs
+++ b/src/webnn_json.rs
@@ -27,6 +27,37 @@ use crate::operators::Operation;
 use std::collections::{BTreeMap, HashMap};
 use webnn_graph::ast::{ConstDecl, ConstInit, GraphJson, Node, OperandDesc};
 
+/// The name used for an operand in the exported AST (and, for constants, as the safetensors tensor
+/// key and `@weights(...)` reference). Falls back to `operand_<idx>` when the operand is unnamed.
+///
+/// Shared by the AST builder and the `.webnn`/`.safetensors` exporter so both agree on the scheme.
+pub(crate) fn operand_export_name(operand: &Operand, idx: usize) -> String {
+    operand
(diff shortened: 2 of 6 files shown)


src/webnn_json.rs
```diff
@@ -209,17 +283,30 @@ pub fn to_graph_json(graph: &GraphInfo, quantized: bool) -> Result<GraphJson, Gr
...
+    match output_override {
+        Some(override_names) => {
+            for (&operand_idx, name) in override_names {
+                if graph.operands.get(operand_idx as usize).is_some() {
+                    outputs.insert(name.clone(), name.clone());
+                }
+            }
+        }
+        None => {
+            for &operand_idx in &graph.output_operands {
```
theHamsta, line 304: You would need to sort output_operands because they were created from a hashmap which returns its contents in a random order.

If doing multiple exports, the output order might be different each time.

GraphJson use BTreeMap which is an ordered dict an therefore preserves order

theHamsta: we could also make our API accept BTreeMap for selecting the outputs and use BTreeMap in GraphInfo where order matters (graph.output_operands is just an array)

src/webnn_json.rs
```diff
@@ -27,6 +27,37 @@ use crate::operators::Operation;
...
+        .clone()
+        .unwrap_or_else(|| format!("operand_{}", idx))
+}
+
+/// Maps operand id -> the name it should be exported under, overriding [`operand_export_name`].
+///
+/// Used by the builder's save path to name graph outputs without mutating the operands: the same
+/// override must apply everywhere the operand is referenced (node inputs, node outputs, and the
+/// graph `outputs` section) so all references stay consistent.
+pub(crate) type OutputNameOverrides = HashMap<u32, String>;
```
theHamsta, line 46: BTreeMap would preserve order of insertion.

Cargo.toml
```diff
@@ -55,6 +55,7 @@ ort = { version = "=2.0.0-rc.12", optional = true, features = [
     "api-24",
 ] }
 half = "2.4"
+safetensors = "0.7"
```
theHamsta, line 59: safetensors just released "0.8" (most changes are for their python bindings).

I realized that it would probably be better long-term to re-export `safetensors` from webnn-graph. Otherwise, when webnn-graph bumps their safetensor version or this repo, we will have this dependency twice (admittedly the dependency is tiny)

Logically, safetensors belongs into `webnn-graph` because it is part of how this format currently stores external weights.

src/webnn_json.rs
mtavenrath: BTreeMap sorts. IndexMap would preserve the order of insertion. Which behavior do we want to have?

theHamsta: I would probably prefer BTreeMap because it is just std library. Predictable behavior would be the main reason for using something different than HashMap. Sorted outputs would be ok for me for the webnn format.

src/webnn_json.rs
mtavenrath: GraphJson,outputs, which is actually the AST, is a BTreeMap and thus ordering is defined to be sorted.

If we'd want to have the original order of insertion we'd have to use an IndexMap.

theHamsta (approved):

PR #189, opened 2026-07-17
fix(trtx): improve trtx engine caching and general cache logic

theHamsta:
## Summary

- turn engine caching for TRT on (so we feel all its bugs)
- invalidate cache key when we edit certain converter files
- hash weights when necessary (non-refittable constants)
- add option to fail on cache failures

The issue that the same graph with constant created in different order has
different hashes persists. JIT caching (helps with engine load times) is still not performed.

## Validation

- [x] `make test`

## Documentation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,7 @@ objc = { version = "0.2", optional = true, features = ["exception"] }
 [build-dependencies]
 prost-build = "0.12"
 cc = "1.0"
+seahash = "4.1.0"
 
 [dev-dependencies]
 tokenizers = { version = "0.22", default-features = false, features = ["onig"] }
--- a/build.rs
+++ b/build.rs
@@ -1,9 +1,11 @@
-#[cfg(feature = "litert-runtime")]
-use std::env;
-use std::fs;
-use std::path::Path;
+use std::hash::Hasher;
+use std::io::Read;
 #[cfg(feature = "litert-runtime")]
 use std::process::Command;
+use std::{env, fs};
+use std::{fs::File, path::Path};
+
+use seahash::SeaHasher;
 
 fn collect_protos(dir: &str) -> Vec<String> {
     let mut files = Vec::new();
@@ -69,6 +71,35 @@ fn build_coreml_protos() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+fn create_source_hash(source_files: &[&str]) {
+    let mut hasher = SeaHasher::new();
+    let mut buffer = Vec::new();
+
+    for file_path in source_files {
+        // Tell Cargo to rerun if ANY of these files change
+        println!("cargo:rerun-if-changed={}", file_path);
+
+        // Only hash the file if it actually exists (prevents crashing if a file is optional)
+        if Path::new(file_path).exists() {
+            let mut file = File::open(file_path).unwrap();
+            buffer.clear(); // Reuse the same buffer to save allocations
+            file.read_to_end(&mut buffer).unwrap();
+
+            // Feed this file's bytes into the running hash state
+            hasher.write(&buffer);
+        }
+    }
+
+    // 2. Finalize the combined hash
+    let total_hash = hasher.finish();
+    let hash_string = format!("{:016x}", total_hash);
+
+    // 3. Write the combined hash out
+    let out_dir = env::var("OUT_DIR").unwrap();
+    let dest_path = Path::new(&out_dir).join("source_hash.txt");
+    fs::write(dest_path, hash_string).unwrap();
+}
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Only compile CoreML protos - ONNX protos come from webnn-onnx-utils
     build_coreml_protos()?;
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -79,6 +79,7 @@ objc = { version = "0.2", optional = true, features = ["exception"] }
 [build-dependencies]
 prost-build = "0.12"
 cc = "1.0"
+seahash = "4.1.0"
```
theHamsta, line 82: we already have this dep as a runtime dependency.

https://crates.io/crates/blake3 seems to be a faster, cryptographic secure and also parallel solution that is as algorithm also known outside of the Rust world. It only works on bytes though, so we would need to replace `Vec` in MLOptions by inline vectors, max fixed length vecs. So that members of GraphInfo are just bytes without any heap allocations.

build.rs
```diff
@@ -98,5 +131,12 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("cargo:rerun-if-changed=protos");
     println!("cargo:rerun-if-changed=build.rs");
     println!("cargo:rerun-if-env-changed=OUT_DIR");
+
+    create_source_hash(&[
+        "src/converters/trtx.rs",
+        "src/converters/trtx_gru.rs",
+        "src/converters/trtx_lstm.rs",
+        "src/converters/trtx_rnn.rs",
+    ]);
```
theHamsta, line 138: I feel we will regret here not including `src/backends/trtx.rs` (e.g. if we ever change build flags for engines), but usually do changes to the backend not affect converted engines.

src/mlcontextoptions.rs
```diff
@@ -85,21 +85,23 @@ pub struct RustNNOptions {
...
 impl Default for TrtxOptions {
     fn default() -> Self {
         Self {
-            // disabled for now, since feature experimental.
-            // can be enabled with more test coverage, but will remain a double-sided sword
-            // e.g. if you change trtx converter, changes might not be visible, since cache skips conversion
-            // maybe the build hash of certain trtx related files could be included in hash
-            engine_caching: false,
-            // Lower CPU overhead using CUDA graphs
+            engine_caching: true,
```
theHamsta, line 102: switching the default... are we confident enough? also a way to find bug: using it.

build.rs
theHamsta: assuming that other backends might also include files here, we will have a single source hash for now. Could also be for each backend to avoid cross-backend invalidation.

src/mlcontextoptions.rs
theHamsta:
```
❯ du -sh  ~/.cache/rustnn/trtx/
1.1G    /home/stephan/.cache/rustnn/trtx/
```
oopsie, maybe turn off until we have some size control logic, like LRU. Again standalone sqlite could be a good solution. As a browser backend, the browser would decide whether to keep cache entries for graphs or not.

theHamsta: most of the engines are just a few hunderd kb big, only some multiple megabytes. Those could come from the `non_refittable_weights` that we would try to avoid completely.

theHamsta: it cuts the time for running all the tests roughly in half. The cache for all the tests is 423MiB big.

running the resnet sample is 13.3s vs 2.7s with caching.

Probably, we would need to compress the engines.


~EDIT: it's indeed the `non_refittable_weights`. Without caching for `non_refittable_weights` (which is mostly the gather tests)~
```
❯ du -sh  ~/.cache/rustnn/trtx/
27M     /home/stephan/.cache/rustnn/trtx/
```

compressed with zip
```
❯ du -sh foo.zip 
2.5M    foo.zip
```

theHamsta: Doing caching for now only if we have no `non_refitable_constants`

theHamsta: I did a mistake and inverted logic to cache only with `non_refitable_constants`

theHamsta: True numbers without `non_refitable_constants`, for the RustNN test suite

uncompressed
❯ du -sh  ~/.cache/rustnn/trtx/       
396M    /home/stephan/.cache/rustnn/trtx/

compressed 
❯ du -sh foo.zip                        
37M     foo.zip
```

Comment (theHamsta) [reply src/mlcontextoptions.rs:102]: resnet50
```
❯ du -sh  ~/.cache/rustnn/trtx/
4.6M    /home/stephan/.cache/rustnn/trtx/

rustnn on  fix-caching [$?] is 📦 v0.5.11 via 🦀 v1.99.0-nightly 
❯ zip foo.zip -r ~/.cache/rustnn/trtx/
updating: home/stephan/.cache/rustnn/trtx/ (stored 0%)
  adding: home/stephan/.cache/rustnn/trtx/54e97e97c98199a2_233_trtx_1.5.0_8d8c63fd2b0d40bc (deflated 91%)
```

theHamsta:
Engine compression: something to add here as a CompressedFileCache or generally useful to have a an optional feature trtx-rs?

The file cache could do streaming compression, trtx would need to do a buffer->buffer compression/decompression. Probably, better here and just tell trtx users how they can make their engines smaller in documentation.

Someone made comparison here https://github.com/hwjsnc/rust-compression-comparison

build.rs
mtavenrath: can't we embed flags into the key for the engine hash?

Cargo.toml
mtavenrath: isn't there a way to reinterpret_cast something to bytes to pass it to blake3?

mtavenrath (dismissed):

theHamsta:
> isn't there a way to reinterpret_cast something to bytes to pass it to blake3?

only if remove `Vec` from MLXXXOptions and replace with something like https://docs.rs/staticvec/latest/staticvec/ that does not use heap allocation and is therefore "just bytes". so a bigger change

build.rs
theHamsta: we could. at the moment, there is only one config.

theHamsta:
https://github.com/rustnn/rustnn/pull/189#discussion_r3614745997

Included the flags into the hash

Issue #191, opened 2026-07-20
Sync bug in trtx?


theHamsta:
I assume the logic around `dispatch` is not 100% correct. I get very rare but all-zero results from time to time.

Need to revise synchronization in dispatch call.

The current idea is that every tensor and every graph have their own CUDA stream and that all the dependencies between the streams need to be captured correctly

I observed the bugs with caching enabled. E.g.
`lstm.https.any.js`

so it could be a missing cache input also

theHamsta:
also `dequantizeLinear.https.any.js` but this was a cuda capture error. Currently, checking whether there was a corrupt jit entry that would have been fixed by #194

theHamsta:
also `reshape.https.any.js`

theHamsta:
`reduce_max.https.any.js`

theHamsta:
Reproduced this on master before caching was enabled `lesser.https.any.js`. but `CUDA_ERROR_STREAM_CAPTURE_INVALIDATED` is related to cuda graphs

theHamsta:
Might be something as simple as they we don't bind the cuda context to thread on dispatch call or that an unsupported op for graph capture is launched


is_infinite.https.any.js
```
[11:52:59.411][2357539][CUDA][E] Returning 901 (CUDA_ERROR_STREAM_CAPTURE_INVALIDATED) from cuLaunchKernel
[11:52:59.411][2357526][CUDA][E] Returning 900 (CUDA_ERROR_STREAM_CAPTURE_UNSUPPORTED) from cuCtxSynchronize_v2
[11:52:59.422][2357529][CUDA][E] Returning 900 (CUDA_ERROR_STREAM_CAPTURE_UNSUPPORTED) from cuCtxSynchronize_v2
[11:52:59.430][2357539][CUDA][E] Capture was invalidated by a prior API call
[11:52:59.430][2357539][CUDA][E] Returning 901 (CUDA_ERROR_STREAM_CAPTURE_INVALIDATED) from cuStreamEndCapture
```
maybe the stream we capture to is used by another operation. We would need an dedicated capture stream

theHamsta:
There seem to be two issues: a sync issue or cuda graph not current issue and cuda graph issue.

[12:09:15.056][3594873][CUDA][E] No CUDA context is current to the calling thread
[12:09:15.056][3594873][CUDA][E] Returning 201 (CUDA_ERROR_INVALID_CONTEXT) from cuCtxGetDevice_v2
[...]

PR #193, opened 2026-07-20
Zstd+Runtime cache

theHamsta:
## Summary

Add global Trtx Runtime cache. #165 but rebased on #192. Running the test suite for trt goes down from 44s from ~250s with engine cache only and ~500s+ without any caching.

~Whole test suite requires `836K` of compressed runtime cache.~ EDIT: running the tests a few times accumulates to 29M compressed 

We use global cache instead of per-engine cache like TRT RTX execution provider does to experiment with things done differently.

Global cache can potentially share cache results between different models, but has disadvantage that it can potentially grow to big without a chance to trim it back. A per-graph cache, would have everything necessary for one graph to reload it again quickly. But all graphs would start with a cold cache after AOT build.

There are some random serialization failures for the cache (don't know yet why this happens). When this happens the JIT cache will not be stored to disk. It will recover automatically the next time the cache is loaded from disk.

## Validation

- [x] `make test`

## Documentation

### DIFF
--- a/src/backends/trtx.rs
+++ b/src/backends/trtx.rs
@@ -719,6 +719,7 @@ impl<'context> MLBackendContext<'context> for TrtxContext<'context> {
             array.as_ptr(),
             array.len(),
         );
+        stream.context().bind_to_thread()?;
         stream
             .memcpy_dtoh(&cuda_tensor.memory, array)
             .to_read_tensor_result(|| tensor.clone())?;
@@ -735,6 +736,7 @@ impl<'context> MLBackendContext<'context> for TrtxContext<'context> {
     ) -> crate::error::Result<()> {
         let cuda_tensor = &mut self.tensors[tensor.id];
         let stream = &cuda_tensor.stream;
+        stream.context().bind_to_thread()?;
         debug!(
             "Uploading tensor {cuda_tensor:?} to array (ptr={:?}, size={:?})",
             array.as_ptr(),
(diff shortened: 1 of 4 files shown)


theHamsta:
This fixed #191 by waiting on the output (not only inputs) on dispatch and by doing cuda graph capture only on second inference.

src/backends/caching.rs
```diff
@@ -144,10 +144,28 @@ pub(crate) fn write_cache_file(cache_path: &Path, content: &[u8]) -> std::io::Re
...
         content.len()
     );
-    let mut file = File::create(cache_path)?;
-    file.lock()?;
-    file.write_all(content)?;
-    Ok(())
+    write_cache_file_atomically(cache_path, |file| file.write_all(content))
+}
+
+fn write_cache_file_atomically(
```
theHamsta, line 150: This is from #194 . This is on top of #194

src/backends/trtx.rs
```diff
@@ -325,6 +400,26 @@ static TRTX_SUFFIX: LazyLock<String> = LazyLock::new(|| {
...
 
+fn jit_cache_disk_key(pci_device_id: &str) -> String {
+    let filesystem_safe_device_id = pci_device_id.replace([':', '.'], "_");
+    format!("{filesystem_safe_device_id}_{}.cache", *TRTX_SUFFIX)
+}
+
+// RuntimeCache owns its TensorRT IRuntimeCache and does not borrow the engine. The lifetime in
+// trtx::RuntimeCache is represented only by PhantomData, so erase it while the cache is held in
+// the process-global map and restore the current engine lifetime when cloning it out.
+fn into_global_runtime_cache<'engine>(
```
theHamsta, line 418: needs to be fixed by trtx

theHamsta: I wanted to change the API for RuntimeCache anyway to not have a Mutex around it (as RuntimeCache is internally holding a mutex already). Would need to do in the next trtx release

mtavenrath (approved):

PR #205, opened 2026-08-13
fest(coreml): Wrap CoreML shim

shubhamg13:
## Summary

- Wrap CoreML shim
 
## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -71,6 +71,24 @@ fn build_coreml_protos() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+#[cfg(all(target_os = "macos", feature = "coreml-runtime"))]
+fn build_coreml_shim() {
+    // On macOS with the CoreML runtime, compile the Objective-C++ exception
+    // firewall (see src/executors/coreml_shim.mm). It catches Objective-C and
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -71,6 +71,24 @@ fn build_coreml_protos() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+#[cfg(all(target_os = "macos", feature = "coreml-runtime"))]
```
theHamsta, line 74: you allow this on Linux (e.g. only guard with coreml-runtime? I regularly check this on Linux. It's currently broken because of the inclusion of 
```
#import <CoreML/CoreML.h>
#import <Foundation/Foundation.h>
```
which I don't have on Linux but otherwise the code path can build on Linux

theHamsta: But you are right, we should wrap it with macos.

I need to fix the shim to work with Linux to be able to remove the macos restriction

shubhamg13: I'm Glad, thanks for merging.

Another thing was bogging me, to consider protos too

Building other platform protos doesn't look good. As it will be waste of resources.

PR #206, opened 2026-08-13
WebNN wasm32 comeback

theHamsta:
## Summary

Bring back WebNN backend by shipping the generated API instead of relying on forked wasm-bindgen.

Let's vendor the generated API until they land in wasm-bindgen.

The WebNN backend or at least the generated bindings could still be their own crate. So, we could do the rustnn, rustnn-cli, webnn-backend split up if we want before. 

There are still some hangs with the tests when I use chromium-canary on Linux.

Also one could wait for async API to create a real WebNN backend implementation. With a real Backend implementation one could also test the Wasm support of the other backends we have (ort, litert,  POC burn)

## Validation

- [ ] `make test`

## Documentation


.github/workflows/ci.yml
```diff
@@ -22,14 +22,20 @@ jobs:
...
       - name: Install Rust test toolchain
         uses: dtolnay/rust-toolchain@1.96.0
+        with:
+         targets: wasm32-unknown-unknown
 
       - name: Install Rust formatting toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt
+          targets: wasm32-unknown-unknown
```
theHamsta, line 28: yes, apparently this is the toolchain we are actually using.

src/backends/webnn/api_generated/gen_Ml.rs
theHamsta, line 1: lots of generated files... Ok, or do we need to have them in some other repo? Would find it need if we could have the feature in RustNN and eventually upstream to wasm-bindgen

theHamsta: The files come from https://github.com/theHamsta/wasm-bindgen/commit/bff148b9cf33582a9a927c0fc721437f21ddd75f where I run the instruction of wasm-bindgen on how to generate new APIs https://github.com/wasm-bindgen/wasm-bindgen/tree/main/crates/web-sys (documented this on the module documentation of `api_generated`

theHamsta:
I know this is not perfect:
- does not test other backends that support wasm32 (ort cpu, ort webgpu, ort webnn, other wasm32 ort backends, litert wasm cpu, literrt webgpu, https://ort.pyke.io/backends/web)
- does not include a backend for WebNN, just a converter
- it can't use the async API because async support is not ready
- other problems with browser, the tests can hang (this is because we don't translate wasm-bindgen JS exceptions into results, web-sys also does not do this, there would be an annotation for this)
- skip list for Linux chromium-canary not complete

But maybe this could be a MVP?

mtavenrath (approved):

PR #208, opened 2026-08-14
deps: bump `ort` dependency to 2.0.0-rc.13

theHamsta:
## Summary

`ort` has a new release that enables new onnxruntime APIs.

After the bump of the `ort` crate, I also ran `cargo update` in a separate commit.

## Validation

- [x] `make test`

## Documentation


theHamsta:
The behavior for missing onnxruntime on Linux is now segfault instead of hang. Seems to come from panicking while other threads are in some C/C++ context.


No changes for WPT tests (I tested with libonnxruntime.so 1.29)

theHamsta:
Eventually, we could try to use `ort::init_from` but I think we would need to be sure that the file we are loading is valid. IMHO, the `ort` crate itself would be a better place to make sure that only valid DLLs and `.so` are loaded and that there is a safe error return if they are not compatible.

mtavenrath (approved):

PR #213, opened 2026-08-28
Stabilize CoreML reduceLogSumExp lowering

matthargett:
## Summary

- [x] Lower CoreML `reduceLogSumExp` with the max-shift identity instead of CoreML's numerically unstable native reduction:
  `max(x) + log(sum(exp(x - max(x))))`
- [x] Preserve WebNN axes, `keepDimensions`, scalar-boundary, and explicit-empty-axes semantics.
- [x] Remove the CoreML expected-failure entries for the WPT overflow and underflow cases.

The reduction keeps dimensions internally so the maximum broadcasts across arbitrary reduced axes, then squeezes or reshapes only at the WebNN output boundary.

## Validation

- [x] `make test-wpt-coreml`: **2438 passed, 44 expected failures skipped, 0 failed** across all 2482 upstream cases. This includes every float32/float16 `reduceLogSumExp` input form plus the `100.0` overflow and `-100.0` underflow cases, with WPT value comparisons.
- [x] `make fmt-check`
- [x] `make docs-backend-ops-check` (report unchanged; the operation was already marked supported)
- [x] Library unit-test phase: **343 passed, 0 failed**, including new lowering-structure coverage for ordinary axes, empty axes, and scalar inputs.
- [x] M4 CoreML `.all` execution compared against a double-precision CPU oracle:
  - WPT stability probes: `[100.0]` and `[-100.0]`, exact
  - CRF normalizer: actual `[101.39999389648438, 97.20000457763672, 101.5]`; expected `[101.4, 97.2, 101.5]`
  - mixture log probability: actual `[99.89554595947266, -99.18876647949219]`; expected `[99.89554546459796, -99.18877002459244]`
- [x] Offline-capable end-to-end compilation on the attached **iPhone XS Max (iPhone11,6), iOS 18.7.9**. The test app contained only the three original WebNN JSON modules plus an arm64 iOS build of RustNN—no `.mlmodel` or `.mlmodelc` artifacts and no network calls. On the phone, each module went through `load_graph_from_path`, `GraphValidator`, and the CoreML converter; the resulting `.mlmodel` was then compiled locally with `MLModel.compileModel(at:)` into a new `.mlmodelc`, loaded, and executed. All three modules produced the same oracle-checked values above and the app exited 0.

This device validation covers input-only modules. On-device packaging and compilation of modules that emit an external CoreML weights blob remains a separate, unvalidated path.

The two model-path probes mirror real community code:

- [`pytorch-crf`'s forward normalizer](https://github.com/kmkurn/pytorch-crf/blob/623e3402d00a2728e99d6e8486010d67c754267b/torchcrf/__init__.py#L238-L260): broadcast score + transition + emission, reduced across source tags.
- [`mixture-density-network`'s loss](https://github.com/tonyduan/mixture-density-network/blob/5bfeae42de2adf42d680bdfee24bb8c2ce52f259/src/blocks.py#L77-L85): component log probability + log mixture weight, reduced across components.

`make test` does not currently complete unmodified on main locally: Rust 1.97 Clippy reports 48 pre-existing diagnostics before tests, and after capping those diagnostics the default no-feature WPT integration binary exits because no runtime backend is enabled. This change adds no Clippy diagnostic; the unit tests and the feature-enabled CoreML WPT suite above pass.

## Documentation

- [x] No user-facing API or backend support-matrix change is needed; this corrects an already-supported CoreML operator.

## CI note

CoreML, both Rust matrix jobs, documentation, and the RustNNPT gate pass. The ONNX and LiteRT WPT jobs are red only on newly fetched `identity → relu` snapshot assertions (ONNX float32/float16; LiteRT float32) introduced by [upstream WPT commit `f0a6949`](https://github.com/web-platform-tests/wpt/commit/f0a694965afd7eb4da71345de936f68d99548dca). Current RustNN `main` has no matching snapshots, and this PR changes neither backend nor the snapshot harness, so those unrelated snapshot additions are intentionally not bundled here.

### DIFF
--- a/src/converters/coreml_mlprogram.rs
+++ b/src/converters/coreml_mlprogram.rs
@@ -2101,6 +2101,233 @@ impl CoremlMlProgramConverter {
         Ok((interleaved_input, interleaved_scale))
     }
 
+    /// Lower `reduceLogSumExp` using the numerically stable max-shift identity:
+    ///
+    /// `log(sum(exp(x))) = max(x) + log(sum(exp(x - max(x))))`.
+    ///
+    /// CoreML's native `reduce_log_sum_exp` overflows and underflows for the
+    /// large-magnitude WebNN conformance inputs. Keep the reduced axes until
+    /// the final step so `max(x)` broadcasts correctly for arbitrary axes.
+    fn emit_stable_reduce_log_sum_exp(
+        graph: &GraphInfo,
+        op: &Operation,
+        overrides: &HashMap<u32, String>,
+        main_block: &mut Block,
+    ) -> Result<(), GraphError> {
+        let (input_id, options) = match op {
+            Operation::ReduceLogSumExp { input, options, .. } => (*input, options.as_ref()),
+            _ => {
+                return Err(GraphError::ConversionFailed {
+                    format: "coreml_mlprogram".to_string(),
+                    reason: "stable reduceLogSumExp lowering called on another operation"
+                        .to_string(),
+                });
+            }
+        };
+        let output_id = op
+            .output_operand()
+            .ok_or_else(|| GraphError::ConversionFailed {
+                format: "coreml_mlprogram".to_string(),
+                reason: "reduceLogSumExp has no output operand".to_string(),
+            })?;
+        let input_operand =
+            graph
+                .operand(input_id)
+                .ok_or_else(|| GraphError::ConversionFailed {
+                    format: "coreml_mlprogram".to_string(),
+                    reason: format!("reduceLogSumExp input operand {} not found", input_id),
+                })?;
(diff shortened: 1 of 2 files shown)


mtavenrath:
It is under discussion by the working group how to handle LogSumExp. Currently the expected test result does not match the spec, which defines LogSumExp as the decomposition of Log, Sum, and Exp. Since exp(100) is inf, the WPT test theoretically should produce inf. See https://github.com/webmachinelearning/webnn/issues/947.

To avoid this overflow, the tendency of the group is to adopt the max technique as proposed in https://github.com/webmachinelearning/webnn/pull/916, so I’m voting to accept this change.

PR #216, opened 2026-08-31
Fix CoreML concurrency

mklimenko-nv:
## Summary

`MLContext::dispatch()` SIGSEGVs on every CoreML-dispatched graph when rustnn is built with any optimization (opt-level >= 1). Debug builds are unaffected, which is why cargo test CI stayed green while downstream consumers running `cargo test --release` (e.g. onnx2webnn's macOS CI) crashed on every test.

## Fix

- Shim: `__bridge_retained`, transferring an owned (+1) reference to the caller.
- Rust: a `ReleaseOnDrop` guard releases the provider after outputs are extracted, on all exit paths, at all three `rustnn_coreml_predict` call sites.

## Validation

- New regression test `backends::coreml::test::coreml_repeated_dispatch_provider_lifetime` — SIGSEGVs (signal 11) with the old shim under `--release`, passes with the fix.
- `cargo test --lib --features coreml-runtime`: 346 passed (debug), 22/22 CoreML tests also pass under `--release`.
- WPT CoreML conformance: 2436 passed, 46 skipped, 0 failed.

### DIFF
--- a/src/backends/coreml.rs
+++ b/src/backends/coreml.rs
@@ -483,4 +483,53 @@ mod test {
         context.read_tensor(&out, &mut result).unwrap();
         assert_eq!(result, &[11, 22, 33, 44]);
     }
+
+    /// Regression guard for the `rustnn_coreml_predict` output-provider lifetime
+    /// bug: the shim returned the prediction result through a plain `__bridge`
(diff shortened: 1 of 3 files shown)


mklimenko-nv:
Onnx2webnn works uncovered more issues than originally estimated, closing this in favor of a combined PR #219

PR #217, opened 2026-09-01
Fix axis input for Squeeze in ONNX

mklimenko-nv:
## Summary

The opset-13+ export path extracted axes only for Unsqueeze; Squeeze fell through to `None`, so the exported node had no axes input. ONNX semantics for an axes-less Squeeze remove all size-1 dims, silently changing the output shape (e.g. [1,128,1,12,64] -> [128,12,64] instead of [1,128,12,64]) while rustnn's internal shape bookkeeping used the correct axes. 

Surfaced as ORT type-inference failures on downstream ops when converting transformers.js models (jina-reranker-v2, marqo-fashionSigLIP).

Squeeze has no WPT coverage (emulation-only, removed from the WebNN spec), so a unit test guards the regression.

### DIFF
--- a/src/converters/onnx.rs
+++ b/src/converters/onnx.rs
@@ -9250,9 +9250,14 @@ impl crate::converters::GraphConverter for OnnxConverter {
                         .static_or_max_shape()
                 });
 
-                // Get axes from typed options (unsqueeze). Typed options required.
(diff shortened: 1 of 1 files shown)


mklimenko-nv:
Closed in favor of combined PR #219

PR #218, opened 2026-09-01
Remove extra fp16 casts

mklimenko-nv:
## Summary

ONNX `Where` export path unconditionally inserted `Cast` nodes on both the true and false value inputs, even when the operand `dtype` already matched `target_type`, producing identity Cast(f16->f16) nodes.

At `GraphOptimizationLevel::Disable` (used by ORT validation sessions), ORT's mandatory `InsertedPrecisionFreeCast` fp16 transform mishandles these identity casts and derives the wrong `dtype` for the  cast output, failing graph load with a `Type Error`.


mklimenko-nv:
Closed in favor of combined PR #219

PR #219, opened 2026-09-01
Onnx2webnn fixes

mklimenko-nv:
## Summary

This PR combines all the fixes necessary on the rustnn side found while enabling transformer.js models in onnx2webnn.

It fixes the bidirectional LSTM/GRU ops, removes extra explicit fp16 case which breaks in optimized onnxruntime builds (#218), fixes axis inputs for Squeeze (#217), as well as CoreML concurrency issues (#216)

### DIFF
--- a/src/backends/trtx.rs
+++ b/src/backends/trtx.rs
@@ -583,7 +583,7 @@ impl<'context, 'builder> MLBackendBuilder<'context, 'builder> for TrtxBuilder<'c
                     let consumers: Vec<String> = graph
                         .operations
                         .iter()
-                        .filter(|op| op.input_operands().contains(id))
+                        .filter(|op| op.all_input_operands().contains(id))
                         .map(|op| op.op_type().to_string())
                         .collect();
                     TrtxError::ConstantRefitFailed {
--- a/src/backends/trtx.rs
+++ b/src/backends/trtx.rs
@@ -279,6 +279,10 @@ impl<'context> TrtxContext<'context> {
         // (mostly scalars)
         config.set_flag(trtx::trtx_sys::BuilderFlag::kREFIT_INDIVIDUAL);
         config.set_flag(trtx::trtx_sys::BuilderFlag::kSTRIP_PLAN);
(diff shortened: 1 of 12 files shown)


src/backends/trtx.rs
```diff
@@ -549,8 +561,29 @@ impl<'context, 'builder> MLBackendBuilder<'context, 'builder> for TrtxBuilder<'c
...
+                    // Name the operand: TensorRT only reports the weight name, and
+                    // "cannot be refitted" means the builder folded the constant.
+                    let consumers: Vec<String> = graph
+                        .operations
+                        .iter()
+                        .filter(|op| op.input_operands().contains(id))
+                        .map(|op| op.op_type().to_string())
+                        .collect();
+                    crate::error::Error::GraphBuildError {
+                        source: crate::GraphError::ConversionFailed {
```
theHamsta: could we try to fade out this generic error and have a dedicated error for this situation? Someone handling the error then also has access to operand causing the error and can match against the kind of error.

mklimenko-nv: Sounds good, added in [13f1ca9](https://github.com/rustnn/rustnn/pull/219/commits/13f1ca96a1d7e79cf4b7510dd6e560292f090597)

src/converters/trtx.rs
```diff
@@ -2996,8 +3127,67 @@ impl TrtxConverter {
             return Ok(());
         }
 
+        // TensorRT (>= 3.16) rejects internal kUINT8 tensors unless a Constant or
```
theHamsta: 3.16? which version does this refer to? Really this old?

theHamsta (dismissed):

src/converters/trtx.rs
mklimenko-nv: @theHamsta, this was a mix-up with the DLA firmware version, removed in [7113bfb](https://github.com/rustnn/rustnn/pull/219/commits/7113bfb1c83afede99c5a0951dc7b883aff77d1d)

theHamsta (approved):

PR #220, opened 2026-09-08
Fix duplicate MLTensor dispatch validation

matthargett:
## Summary

- Validate duplicate `MLTensor` dispatch bindings by tensor identity rather than binding name.
- Reject reuse of the same logical tensor across input and output bindings before backend dispatch.
- Update existing dispatch tests so successful and unrelated-error cases use distinct input and output tensors.

The previous map stored `binding name -> tensor id`. Because `MLNamedTensors` already enforces unique names, that check could not detect one tensor bound under different names. This changes the map to `tensor id -> first binding name` and adds backend-independent regression coverage.

## Standards and upstream test coverage

This implements the `MLContext.dispatch()` requirement that the combined input and output tensor list contain no duplicate items.

Chromium currently checks repeated output tensors and input/output reuse by `MLTensor` object membership in [`ValidateMLTensorUsage()`](https://github.com/chromium/chromium/blob/a4224948aecb4020c0efbc1936be667fa1a17fba/third_party/blink/renderer/modules/ml/ml_context.cc#L296-L318). Its [`WebNNGraphDispatchTest`](https://github.com/chromium/chromium/blob/a4224948aecb4020c0efbc1936be667fa1a17fba/third_party/blink/renderer/modules/ml/webnn/ml_graph_test.cc#L1209-L1267) exercises successful repeated dispatch, but I could not find an alias-rejection assertion there. I also could not find a corresponding dispatch/`MLTensor` alias case in the current WPT [`webnn/validation_tests`](https://github.com/web-platform-tests/wpt/tree/master/webnn/validation_tests) directory or another conformance suite in the WebNN standards repository.

Would you all prefer a follow-up WPT validation test covering duplicate inputs, duplicate outputs, and input/output reuse, or is there another WebNN conformance suite where these cases belong? (or is this not really a problem?)

## Validation

- `make fmt-check`
- Focused regression: 1 passed, 0 failed
- `cargo test --lib`: 341 passed, 0 failed
- `make docs-backend-ops-check`

`make test` reaches the existing Rust 1.97 Clippy gate and stops on 48 diagnostics in unchanged files. The changed file introduces no new diagnostic; formatting, focused coverage, the complete library suite, and the documentation drift check pass.

### DIFF
--- a/src/mlcontext.rs
+++ b/src/mlcontext.rs
@@ -31,6 +31,23 @@ pub type MLNamedTensors<'names> = BTreeMap<&'names str, &'names MLTensor>;
 /// <https://www.w3.org/TR/webnn/#typedefdef-mlnamedoperands>
 pub type MLNamedOperands<'names> = BTreeMap<&'names str, MLOperand>;
 
+fn validate_unique_tensor_bindings(
+    inputs: &MLNamedTensors,
+    outputs: &MLNamedTensors,
+) -> Result<()> {
+    let mut all_tensor_ids = HashMap::new();
+    for (&name, &tensor) in inputs.iter().chain(outputs.iter()) {
+        if let Some(other_name) = all_tensor_ids.insert(tensor.id, name) {
+            return Err(Error::DuplicateTensorBinding {
+                aliased_tensor: tensor.clone(),
+                first_binding: other_name.to_string(),
+                other_binding: name.to_string(),
+            });
+        }
+    }
+    Ok(())
+}
+
 pub use crate::mlgraphbuilder::MLGraphBuilder;
 use crate::{
     backend_selection::{select_backend, select_backend_by_gpu},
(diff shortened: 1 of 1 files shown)


theHamsta (dismissed):

src/mlcontext.rs
```diff
@@ -31,6 +31,23 @@ pub type MLNamedTensors<'names> = BTreeMap<&'names str, &'names MLTensor>;
 /// <https://www.w3.org/TR/webnn/#typedefdef-mlnamedoperands>
 pub type MLNamedOperands<'names> = BTreeMap<&'names str, MLOperand>;
 
+fn validate_unique_tensor_bindings(
+    inputs: &MLNamedTensors,
+    outputs: &MLNamedTensors,
+) -> Result<()> {
+    let mut all_tensor_ids = HashMap::new();
+    for (&name, &tensor) in inputs.iter().chain(outputs.iter()) {
```
theHamsta, line 39: Hi! could you also add a test that aliases a name between input and output? because I think this situation does not get validated with the new solution.

inputs = {"a": tensor_a}; output = {"a": tensor_b}

matthargett: Added the inverse case in 6427751: matching names across the input and output maps remain valid when they refer to distinct tensors, consistent with the draft's identity-based duplicate check.

theHamsta (approved):

PR #221, opened 2026-09-08
tests: do API trace for WPT tests when TRT_SHIM_OUTPUT_JSON_FILE is set

theHamsta:
## Summary

## Validation

- [ ] `make test`

## Documentation

### DIFF
--- a/tests/run_wpt_conformance.rs
+++ b/tests/run_wpt_conformance.rs
@@ -118,6 +118,14 @@ fn main() {
     let args = Arguments::from_args();
     let wpt_dir = default_wpt_dir();
 
+    // See https://docs.nvidia.com/deeplearning/tensorrt-rtx/latest/inference-library/capture-replay.html#capture-replay
+    // for this Linux-only feature that is enabled via TRT_SHIM_OUTPUT_JSON_FILE
+    #[cfg(all(unix, feature = "trtx-runtime"))]
+    if std::env::var("TRT_SHIM_OUTPUT_JSON_FILE").is_ok() {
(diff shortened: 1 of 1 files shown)


tests/run_wpt_conformance.rs
```diff
@@ -118,6 +118,14 @@ fn main() {
     let args = Arguments::from_args();
     let wpt_dir = default_wpt_dir();
 
+    // See https://docs.nvidia.com/deeplearning/tensorrt-rtx/latest/inference-library/capture-replay.html#capture-replay
+    // for this Linux-only feature that is enabled via TRT_SHIM_OUTPUT_JSON_FILE
+    #[cfg(all(unix, feature = "trtx-runtime"))]
+    if std::env::var("TRT_SHIM_OUTPUT_JSON_FILE").is_ok() {
+        trtx::dynamically_load_tensorrt(Some("libtensorrt_shim.so"))
+            .expect("Failed to load libtensorrt_shim.so for TensorRT API capture");
```
theHamsta, line 126: this works, but of course makes only really sense when you also filter for a single test.

e.g. 
```
export TRT_SHIM_OUTPUT_JSON_FILE=/tmp/foo.json
export LD_LIBRARY_PATH=~/Downloads/TensorRT-RTX-1.6.1.120/lib
cargo t -F trtx-runtime argMin_float32_4D_tensor__axis_0__all_options

```

would generate `/tmp/foo.json` and `/tmp/foo.bin`

theHamsta: hmm, there is also a segfault :shaking_face:

theHamsta: only with one TRT version

Issue #222, opened 2026-09-08
Track SmolLM-135M CoreML blockers


matthargett:
## Scope

Reproduced with the SmolLM-135M WebNN export against #219 at `13f1ca9` on an M4 Mac.

The graph contains 26 distinct operations. Of those, the CoreML converter already recognizes 25; the only initially unimplemented operation was the internal `shape` interchange op, used 127 times. The recognized set is: `add`, `cast`, `concat`, `cos`, `div`, `equal`, `expand`, `gather`, `greater`, `matmul`, `mul`, `neg`, `pow`, `reduceMean`, `reshape`, `scatterND`, `sigmoid`, `sin`, `slice`, `softmax`, `sqrt`, `transpose`, `triangular`, `unsqueeze`, and `where`.

## Checklist

- [x] Lower internal `shape` nodes to MIL `shape` with CoreML's int32 proxy for WebNN int64 shape tensors (`src/converters/coreml_mlprogram.rs`) — #223.
- [x] Preserve shape-tensor type and rank through `unsqueeze` during WebNN text import (`src/webnn_json.rs`) — #223.
- [x] Cast integer `where` conditions to uniquely named MIL bool values so comparison outputs are not redefined (`src/converters/coreml_mlprogram.rs`) — #223.
- [x] Emit CoreML shape ranges for dynamic graph inputs/outputs and pass the active concrete tensor shape at dispatch.
- [x] Preserve dynamic shapes through `scatterND`, `slice`, and `triangular`, and distinguish scalars from zero-extent tensors.
- [ ] Define the first-token zero-length KV-cache behavior. The proposed WebNN semantics are tracked in [webmachinelearning/webnn#391](https://github.com/webmachinelearning/webnn/issues/391); CoreML flexible-shape ranges still have a minimum extent of 1.
- [x] Bisect the full SmolLM graph's nonzero dynamic-resize failure to maximum extents introduced by synthesized CoreML helper tensors. Gather index normalization expanded active `[1, 1]` indices to `[8, 4096]`; preserving its dynamic MIL dimensions reduces the embedding/projection result from 75,497,472 bytes to the expected 2,304 bytes — [fork PR #9](https://github.com/rebeckerspecialties/rustnn/pull/9).
[...]

PR #223, opened 2026-09-08
Support SmolLM shape tensors in CoreML

matthargett:
## Summary

- lower internal `shape` nodes to MIL `shape`
- infer and preserve shape-tensor rank and type through imported `unsqueeze` nodes
- cast integer `where` conditions to uniquely named MIL bool values
- verify exact CoreML predictions for `shape -> unsqueeze` (int64 and uint32 readback) and zero/nonzero integer masks

This completes the first three items in #222. The SmolLM compiler pass now reaches the dynamic `attention_mask` input-signature mismatch tracked there.

## Validation

- M4, Rust 1.97.0: `make fmt-check`, `make lint`, and strict Clippy with `coreml-runtime,dynamic-inputs` pass
- Rust library tests: 357 default; 364 with `coreml-runtime`; 371 with `coreml-runtime,dynamic-inputs`; all pass
- TensorRT option-operand consumer regression test passes
- full CoreML WPT at `04cd4a830e26add75d385f28d95a080600b5d70f`: 2,460 passed, 41 expected skips, 20 existing negative-index scatter failures. Results match a clean main control case-for-case
- `make docs-backend-ops-check` passes
- earlier M4 SmolLM-135M validation: compilation advances from unsupported `shape` and duplicate MIL values to `Model input 'attention_mask' has a different shape than its corresponding parameter to main.`

The public builder/interchange shape dtype remains int64; the uint32 test uses an explicit cast. This does not implement #225's proposed dynamic-shape API. A reduced ranged `shape -> unsqueeze` probe still reproduces the input-signature mismatch tracked in #222; the passing numerical tests use fixed dimensions and repeated CPU-only predictions.


mtavenrath (approved):

mtavenrath:
@matthargett  there is a minor ci failure: /home/runner/work/rustnn/rustnn/docs/development/backend-operator-support.md is out of date. Run make docs-backend-ops and commit the result. Can you update the file?

PR #226, opened 2026-09-11
Add a workflow to create a sync PR to update WPT SHA+snaps+expectations.

shubhamg13:
## Summary

  - Adds a workflow which syncs the WPT expectations by:
    - Updating the test snapshots.
    - Updates the expected failures.
    - And creates an auto-PR .
  - This workflow is scheduled to trigger on Monday 03:00 UTC.
  - Apart from scheduled run, it can also be manually run.

This saves from failures from update in upstream WPT corpus while merging new PRs.

## Validation
- [ ] `make test`

  - Result from Action Run in `main` fork: https://github.com/shubhamg13/rustnn/actions/runs/34806399784
  - PR created as an output for Action Run: https://github.com/shubhamg13/rustnn/pull/5

## Documentation


shubhamg13:
I think it will be good addition. If, yes: Can we force push this and run it to do the job.

tests/run_wpt_conformance.rs
```diff
@@ -98,12 +98,8 @@ fn push_backend_trials(
...
                     let msg = err.message().unwrap_or("test failed");
-                    if backend_prefix != "coreml" {
-                        insta::assert_snapshot!(
-                            snapshot_name,
-                            format!("{file_name} {test_name}, {backend_prefix}\n {msg}")
-                        );
-                    }
+                    // Failures are tracked in {backend}_expected_failures.txt, not
+                    // snapshotted: snapshots only pin passing state (see below).
                     report.record_fail(&file_name, &test_name, &backend_prefix, msg, duration);
```
theHamsta, line 103: why do we then need snapshots at all? The idea of snapshots was mainly to monitor failures.

Ok for me if we only have expected failure list. Apparently, one effect of the snapshots is still kept: we still fail when we actually pass an expected failure which would require removing this test from the failing list.

theHamsta:
Worked great for me locally, though we add new code for scripts.

With the updates we would get 10 new failures for TRT. @mtavenrath on whether we want to fix those or just also have automated expected failures for trt.

All failures are related to negative indices of scatterND

theHamsta (approved):

.github/workflows/snapshot-sync.yml
```diff
@@ -0,0 +1,159 @@
...
+  workflow_dispatch:
+
+concurrency:
+  group: wpt-snapshot-sync
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+  pull-requests: write
+  actions: write
```
theHamsta, line 16: why is the `actions: write` permission needed?

theHamsta (changes requested):
Wouldn't it be good to pin the commit of the WPT repo (I can't find where the WPT test repo is pinned. Could it be just a `git submodule` or checked it known good commit hash? )? This way we can have passing CIs until we get the automated test update PR.

Otherwise, we would need to manually trigger this workflow whenever we have failing CI

theHamsta (dismissed):
Would be also fine to merge right away and pin the WPT suite in a follow up to unblock other PRs.

shubhamg13:
> Wouldn't it be good to pin the commit of the WPT repo (I can't find where the WPT test repo is pinned. Could it be just a `git submodule` or checked it known good commit hash? )? This way we can have passing CIs until we get the automated test update PR.
> 
> 
> 
> Otherwise, we would need to manually trigger this workflow whenever we have failing CI 

My idea is to have a weekly sync. In case something lands in between we can trigger a manual sync.



If I understand you correctly, you want to trigger it in every  merge request on WPT

- Hmm, not tracking it. Using existing fetch-wpt command to get latest corpus.

shubhamg13:
> Would be also fine to merge right away and pin the WPT suite in a follow up to unblock other PRs.

If I remember correctly, when merging, authors do a main branch merge.

If that is the case then there will be one more additional step for authors to trigger workflow before that. Right?

theHamsta:
> > Would be also fine to merge right away and pin the WPT suite in a follow up to unblock other PRs.
> 
> If I remember correctly, when merging, authors do a main branch merge.
> 
> If that is the case then there will be one more additional step for authors to trigger workflow before that. Right?

My idea was to always test on a fixed revision of the WPT tests even if that revision is not the newest. Your automated PR would sync that revision to the newest revision of WPT, potentially bringing new updates.

This way we can separate changes to our code base from changes to WPT.

Similar like we have `Cargo.lock` that tracks git deps, we only get updates when we run `cargo update`

shubhamg13:
> > > Would be also fine to merge right away and pin the WPT suite in a follow up to unblock other PRs.
> > 
> > 
> > If I remember correctly, when merging, authors do a main branch merge.
> > If that is the case then there will be one more additional step for authors to trigger workflow before that. Right?
> 
> My idea was to always test on a fixed revision of the WPT tests even if that revision is not the newest. Your automated PR would sync that revision to the newest revision of WPT, potentially bringing new updates.
> 
> This way we can separate changes to our code base from changes to WPT.
> 
> Similar like we have `Cargo.lock` that tracks git deps, we only get updates when we run `cargo update`

I see, I came to this idea as I encountered failures recently when WPT upstream was changed and we had stale snapshots. 
And similar weekly workflow is at Servo and Chrome also (but not sure about the frequency)

I think we can add another workflow to verify/run with existing corpus (or perhaps extend it with some inputs)

Let me know, what you think.

mtavenrath:
pinning the WPT tests and updating them frequently seems like the most stable way to stabilize our CI. Updating the tests will result in a PR showing us what to fix for new or updated tests. The only case which might get ugly with this workflow would be tests changing their result.

shubhamg13:
> pinning the WPT tests and updating them frequently seems like the most stable way to stabilize our CI. Updating the tests will result in a PR showing us what to fix for new or updated tests. The only case which might get ugly with this workflow would be tests changing their result.

_Maybe I'm misunderstund something here._ 

Pinning the WPT to a specific commit, got it.


Updated, please take a look.

tests/run_wpt_conformance.rs
shubhamg13: Initially, I thought same. 

But wasn't sure that snapshots serve any other purpose or not. 

Give me green light I can clean all snapshots in next PR. I Can modify this script as well.

.github/workflows/snapshot-sync.yml
shubhamg13: Updating: (Residue from initial idea, Initially planned to auto commit)

scripts/fetch_wpt.mjs
```diff
@@ -2,13 +2,16 @@
 /**
  * Shallow sparse-clone or update the Web Platform Tests repository for WebNN conformance.
  *
+ * The checked-out revision is pinned by the `WPT_REVISION` file at the repo root,
+ * so CI always tests against a known-good WPT commit. The scheduled sync workflow
```
shubhamg13: TODO: Add a sync workflow after.

tests/run_wpt_conformance.rs
theHamsta: Ok for me to remove the snapshots.

Their main purpose was to know whether a weird failure is the behavior of the backend (for all targeted devices) or a weird local issue + detect changes in behavior (different kind of test failure, change from failure to passed test).

It would be nice if it would be easy bring them back for local debugging. E.g. by keeping the insta dev dependency. just out-commenting the snap-shot statements, opt-in to snap-shotting via env var. I used snap shots to compare results on different platforms (devices, OSes, inter-run variance of failures)

scripts/fetch_wpt.mjs
```diff
@@ -19,10 +22,22 @@ const repoRoot = path.resolve(__dirname, '..');
 const cacheDir = path.join(repoRoot, '.cache');
 const wptDir = process.env.WPT_DIR ?? path.join(cacheDir, 'wpt');
 const repo = 'https://github.com/web-platform-tests/wpt.git';
+const revisionFile = path.join(repoRoot, 'WPT_REVISION');
```
theHamsta, line 25: as far I can see is WPT_REVISION only checked during `wpt-fetch` and `wpt-sync`. When, I would just run WPT without syncing I would risk that I'm running an outdated  WPT revision: a version that does not match the pinned version.

Could we check during `cargo t` that the WPT repo matches the WPT_REVISION pinned file? Or at least warn on a mismatch?

theHamsta: On CI, we should always be fine by using `wpt-fetch` always.

theHamsta:
@shubhamg13 apparently there is still a problem with permissions: https://github.com/rustnn/rustnn/actions/runs/34849977598

shubhamg13:
> @shubhamg13 apparently there is still a problem with permissions: https://github.com/rustnn/rustnn/actions/runs/34849977598

_I encountered the same scenario while testing._ 

We need to resolve this by granting access to GitHub Actions from repository `Settings -> Actions -> General`. 

Screenshot 2026-09-14 at 10 03 11 PM


Can you please check once.

PR #228, opened 2026-09-14
Fix CoreML gather bounds for active dimensions

matthargett:
`gather`, `gatherElements`, and `gatherND` normalize indices against the declared maximum data extent, not the active extent. With a maximum of 8 and an active length of 2, `-1` becomes 7 instead of 1. This can return incorrect values as dynamic inputs grow or shrink.

Read indexed dynamic dimensions through MIL `shape` and `slice_by_size`, then use those bounds for wrapping and clamping. Static indexed axes retain constant bounds, including when unrelated axes are dynamic. Keep the CoreML7 target and reject dynamic `gatherND` tuple widths instead of silently using their maximum.

Scalar indices stay as one-element vectors through normalization and gathering. An explicit squeeze removes only the indexed axis, preserving dynamic extents and unrelated singleton dimensions. This avoids the native scalar-index gather path that returned no output in macOS CI. Scalar results retain CoreML's `[1]` boundary representation without changing the WebNN descriptor.

This addresses the active data-axis follow-up from [#227](https://github.com/rustnn/rustnn/pull/227), related to [#222](https://github.com/rustnn/rustnn/issues/222).

Regression coverage reuses compiled graphs across growing/shrinking extents, checks constant/runtime negative and out-of-range indices, and verifies actual output shapes. Additional cases cover scalar indices on first/middle/last axes, rank-five inputs, and singleton dimensions. `make test-coreml-gather` runs the focused checks; main's broader CoreML build/test targets remain intact.

Validation on M4 Max/macOS 27:

- Formatting and strict Clippy, including CoreML with and without dynamic inputs.
- `make test`; all-target CoreML builds and library/integration tests with both feature configurations; all 19 focused gather checks pass.
- CoreML WPT: 69 gather cases pass with one existing skip. Full suite: 2,470 passed, 60 existing skips, no failures, using [WPT `9f84649`](https://github.com/web-platform-tests/wpt/commit/9f84649df7a12516b52441a7660dc36b8837a602). No tolerances or expected-failure lists changed.
- Strict MkDocs, rustdoc, and backend operator-report checks pass.

The original active-index bounds reproducer was also validated on iPhone XS Max/iOS 18.7.10: 20 cases under CPU-only and `.all`, with exact shapes and zero numerical error, compiling the portable source locally. That device run predates the scalar-index CI follow-up.


mtavenrath:
@matthargett Two tests are failing and there is a merge conflict.

[2](https://github.com/rustnn/rustnn/actions/runs/35840577371/job/107114392595?pr=223#step:22:43)

failures:
    gather_reports_actual_scalar_and_dynamic_nonindexed_output_shapes
    scalar_index_gather_preserves_dynamic_nonindexed_output_on_dispatch

matthargett:
Thanks. Both failures report a missing `result` at `rows=4`, rather than a numerical mismatch; the checked path fails even with CPU-only. I’ll resolve the conflict with main and address the missing-output path while preserving the active-dimension bounds and exact shape/value checks, then rerun fmt, clippy, and the CoreML tests locally before pushing.

matthargett:
Fixed in [0461a7b](https://github.com/rustnn/rustnn/commit/0461a7b45db197602db3eab268b2a01445ace9e0). I merged main and kept scalar indices as one-element vectors through gather, then squeezed only the indexed axis. Both reported tests now pass on the macOS 26.6.2 CI runner, with exact shape/value checks intact. Added coverage for singleton dimensions, rank-five inputs, and first/middle/last axes. Local fmt, strict clippy, CoreML tests, and full CoreML WPT pass; all CI checks are green.

PR #229, opened 2026-09-14
feat(cann): run WPT conformance on-device

shubhamg13:
## Summary

Register the CANN/HiAI backend with the WPT harness and add the plumbing to run the conformance corpus on the NPU over hdc:

  - register the "cann" WPT backend
  - embed the WPT corpus at build time (`wpt-embed-corpus`; the build script runs
    Node on the host), so no Node.js is needed on the device
  - run every trial (no skips); baseline failures in cann_expected_failures.txt,
    fp16 tolerance floor
  - test-wpt-cann/wpt_on_device (push binary+DDK, run, pull JSON report);
    wpt-sync-cann regenerates the baseline
  - drop the converter's supported-op gate; fix the output-index panic it hid

## Validation

- [ ] `make test`

## Documentation


tests/wpt_conformance/wpt_js_loader.rs
```diff
@@ -132,6 +133,18 @@ pub fn check_wpt_revision(wpt_dir: &Path) -> Option<String> {
 
 /// Load the full WPT conformance corpus in one Node.js invocation.
 pub fn load_wpt_corpus(wpt_dir: &Path) -> Result<WptCorpus, String> {
+    // On-device runs (e.g. CANN) have no Node.js; load a pre-dumped corpus JSON
+    // from the host instead of spawning the Node bridge.
+    if let Ok(path) = std::env::var("WPT_CORPUS_JSON")
```
theHamsta: for WebNN (target wasm32) I'm embedding the test corpus inside the test binary https://github.com/theHamsta/rustnn/blob/5c095d678125dfdd2db0e04b8b0f1d5f0adae513/build.rs#L120 . So, node runs during build and is then embedded into the test binary.

`const WPT_CORPUS: &str = include_str!(concat!(env!("OUT_DIR"), "/webnn-wpt-corpus.json"));`

Would this work also for CANN? Or are you building on device? If you're using cross-compilation, then your host can run node and the target will have the JSON embedded

shubhamg13: I'm doing cross compilation for OHOS though. 

I will then wait for your solution. And check if it is possible to run.

shubhamg13: Thanks for the insight. I have updated it to embed wpt corpus.

theHamsta (approved):

theHamsta:
Thank you!

PR #233, opened 2026-09-16
Add experimental C/C++ bindings

theHamsta:
## Summary

Generate C API header via cargo-c (https://github.com/lu-zero/cargo-c) and add a C++ header that uses the C header.

## Validation

- [ ] `make test` (I tested all new features)

## Documentation


.github/workflows/ci.yml
```diff
@@ -137,6 +137,35 @@ jobs:
...
+      - uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@1.97.0
+
+      - name: Install protobuf compiler and cargo-c
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y protobuf-compiler
+          cargo install cargo-c
```
theHamsta: this takes a long time. should cache or download. I didn't trust the external install action that downloaded it for me automcatically

theHamsta: brew is faster, but still a bit slow compared to just getting the binary.

theHamsta: 1min30s for whole workflow might be ok though.

assets/capi/include/rustnn.hpp
```diff
@@ -0,0 +1,493 @@
...
+                                  named_inputs.size(), named_outputs.data(),
+                                  named_outputs.size()));
+  }
+
+private:
+  friend class GraphBuilder;
+  RustnnContext *value_ = nullptr;
+};
+
+class GraphBuilder {
```
theHamsta: The API currently only covers some ops. Seeing how much more code it would be for the remaining ops. Seeing how much more it would be for all ops.

assets/capi/include/rustnn.hpp
```diff
@@ -0,0 +1,493 @@
...
+};
+
+inline void check(RustnnStatus status) {
+  if (status != RustnnStatus_Success) {
+    const char *message = rustnn_last_error_message();
+    throw Error(message != nullptr ? message : "unknown rustnn error");
+  }
+}
+
+enum class DataType {
```
theHamsta: TODO: keep `ML` prefix for C and C++ API

assets/capi/include/rustnn.hpp
theHamsta: Tried to add remaining ops.

assets/capi/include/rustnn.hpp
```diff
@@ -0,0 +1,756 @@
...
+#include <optional>
+#include <stdexcept>
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace rustnn {
+
+inline void initializeLogger() { rustnn_init_logger(); }
```
mtavenrath: is there a way to auto-generate doxygen style documentation for the api or at least make generate it based on the bikeshed file?

mtavenrath: can this be called if pretty_env_logger is optional and disabled?

Cargo.toml
```diff
@@ -226,5 +227,64 @@ required-features = ["native-examples"]
 
 [lib]
 name = "rustnn"
-crate-type = ["rlib"]
+crate-type = ["rlib", "cdylib", "staticlib"]
```
mtavenrath: can be make cdylib and staticlib optional to avoid increasing build time when developing only on the rust side?

examples/capi/README.md
```diff
@@ -0,0 +1,268 @@
+# rustnn C and C++ examples
+
+## TL;DL
```
mtavenrath: TL;DR?

mtavenrath (commented):
should we document multithreading behavior in general?

assets/capi/include/rustnn.hpp
theHamsta: Well, it is not optional for this config. I know pretty_env_logger is not optimal but you would need some kind of access to the logging system. A callback-based logger might also be nice future feature.

We can think of optional and not optional features of RustNN when we have split out the `main.rs` to its own `rustnn-cli` crate.

examples/capi/README.md
theHamsta: aka, this part of the docs is not AI generated and all you need as a human reader. This part is my summary of the C:stuck_out_tongue:

PR #235, opened 2026-09-17
Fix CoreML triangular masking and diagonal handling

matthargett:
Fix CoreML `triangular` masking and diagonal handling:

- Use nonnegative band widths for upper triangles with negative diagonals; CoreML interprets negative widths as unlimited.
- For floating-point triangles excluding the main diagonal, build an active-shape mask with `fill_like` and select values instead of subtracting. Masked NaN/infinities become zero, and retained values stay unchanged.
- Preserve integer subtraction for the corresponding int32 cases and use URL compilation for those graphs. This preserves the tested large integers on the M4/macOS 27, but is not a portable fix for CoreML's integer precision loss.

Band widths are bounded safely, including `i32::MIN`, without freezing active dimensions. Shared CoreML build/test targets cover unit and integration tests with dynamic inputs enabled and disabled; WPT remains separate.

Validation:

- M4/macOS 27: formatting, strict all-target Clippy, 359 default, 368 CoreML, and 381 CoreML/dynamic library tests pass. CoreML builds and runtime suites pass with dynamic inputs both disabled and enabled. The temporarily ignored large-int32 regression also passes when explicitly invoked with `--ignored` on this machine.
- Eight [WPT additions](https://github.com/web-platform-tests/wpt/pull/62756) fail before and pass after on the M4. The complete triangular file passes 42/42 with zero observed error.
- iPhone XS Max, iOS 18.7.10: 12 uncompiled model sources compile locally; all 24 floating-point cases pass exact shape/value checks. Final floating-point model emission matches these tested sources after protobuf decoding. This did not validate the large-int32 cases.

Known limitation: native CoreML loses int32 precision on the macOS 26.6.2 runner and physical A12/iOS 18.7.10 and A10X/tvOS 26.6 devices. [Standalone probes](https://github.com/rebeckerspecialties/rustnn/blob/diagnostics/triangular-int32/diagnostics/README.md) reproduce this without RustNN; allowing all compute units does not fix it. Following [maintainer guidance](https://github.com/rustnn/rustnn/pull/235#issuecomment-5790829434), only `triangular_int32_keeps_values_beyond_float32_precision` is temporarily ignored. Its exact assertions are unchanged and its explicit invocation is documented. Ordinary int32, floating-point, and dynamic-shape tests remain active; a portable large-int32 fix is still needed.

Related to #222.


Makefile
```diff
@@ -175,6 +175,9 @@ test-wpt-litert:
 test-wpt-coreml:
 	$(CARGO) test --test run_wpt_conformance --features coreml-runtime -- coreml --test-threads 1
 
+test-coreml-triangular:
+	$(CARGO) test --test test_coreml_triangular --features coreml-runtime,dynamic-inputs -- --test-threads=1
```
theHamsta, line 179: why have this as a separate target in makefile and CI? Couldn't we just always run  `--features coreml-runtime,dynamic-inputs` tests or have the dynamic variant tested additionally? The makefile or just `cargo t` can already filter if you just request `trianglular`

matthargett: Agreed. I added it because the existing CoreML CI step uses `--lib`, which would skip these integration tests. A shared CoreML integration-test run with `coreml-runtime,dynamic-inputs` makes more sense than a target per op, lmk if you'd like me to add that small infra into this existing PR :D

theHamsta: sure, feel free to add the testing needed here if it is not too much work. More testing is probably always better.

matthargett: replaced the triangular-only target with shared CoreML build and test targets. so CI now covers unit and integration tests with dynamic inputs both enabled and disabled :D I kept WPT separate. on my local M4 macOS 27, both configurations and all 42 triangular WPT cases pass locally (which includes the new tringular tests I submitted to WPT).

matthargett:
ok, this is weird (to me anyway) about the CI failures:
Native CoreML on the macOS 26.6.2 runner returns 0 for 16,777,217 − 16,777,216; the M4/macOS 27 returns 1. This reproduces without RustNN. Exact input readback, explicit output buffers, alternate operator sets, and several alternative lowerings rule out those workarounds.

I cross-checked this on physical A10X/tvOS 26.6 and A12/iOS 18.7.10 devices using standalone Swift/CoreML probes, compiling the models on each device. Both consistently lose int32 precision—even 16777217 - 16777216 returns 0 instead of 1 with CPU-only execution. (Expanding it to .all doesn't magically fix things).

This isn’t confined to GitHub runners or macOS 26.6.2. The subtraction-based lowering passes on my M4/macOS 27, but fails on both older devices, so it isn’t a portable workaround.

I _think_ we should keep the exact-value assertion intact? We still need an exact lowering or an explicit unsupported-path error, successful compilation shouldn’t silently produce incorrect integers, right @theHamsta ? I have some ideas on how to do the fallback but want to double check on alignment (har har) and goals first.

theHamsta:
> ok, this is weird (to me anyway) about the CI failures: Native CoreML on the macOS 26.6.2 runner returns 0 for 16,777,217 − 16,777,216; the M4/macOS 27 returns 1. This reproduces without RustNN. Exact input readback, explicit output buffers, alternate operator sets, and several alternative lowerings rule out those workarounds.
> 
> I cross-checked this on physical A10X/tvOS 26.6 and A12/iOS 18.7.10 devices using standalone Swift/CoreML probes, compiling the models on each device. Both consistently lose int32 precision—even 16777217 - 16777216 returns 0 instead of 1 with CPU-only execution. (Expanding it to .all doesn't magically fix things).
> 
> This isn’t confined to GitHub runners or macOS 26.6.2. The subtraction-based lowering passes on my M4/macOS 27, but fails on both older devices, so it isn’t a portable workaround.
> 
> I _think_ we should keep the exact-value assertion intact? We still need an exact lowering or an explicit unsupported-path error, successful compilation shouldn’t silently produce incorrect integers, right @theHamsta ? I have some ideas on how to do the fallback but want to double check on alignment (har har) and goals first.

I also observed differences between CI and local machines. Therefore, the automated update of expected_failures is problematic. We would also need manual expected failures to account for locally observed differences, not present in CI.

Apparently, `macos-latest` is MacOS26 (https://github.com/actions/runner-images). The new MacOS27 is not yet available. 

> I cross-checked this on physical A10X/tvOS 26.6 and A12/iOS 18.7.10 devices using standalone Swift/CoreML probes, compiling the models on each device. Both consistently lose int32 precision—even 16777217 - 16777216 returns 0 instead of 1 with CPU-only execution. (Expanding it to .all doesn't magically fix things).

You could add a `#[ignore = "6777217 - 16777216 returns 0 on some configurations including CI"]` to the test until we know of a better solution.

matthargett:
done in f682f6c. I ignored only the large-int32 regression, kept its exact assertions, and documented how to run it explicitly. The other triangular tests remain active. Local fmt, strict Clippy, both CoreML configurations, and all 42 triangular WPT cases pass; CI is now green too. The underlying CoreML precision issue still needs a portable fix.

theHamsta:
Thanks! And please excuse any delays in reviewing!

PR #237, opened 2026-09-18
feat(cann): Expand CANN coverage (I)

shubhamg13:
## Summary

  - compile and load the model at build() (hiai-rs rev a7b8ae5)
  - broaden WebNN op coverage and NCHW/int32 handling
  - own the DDK graph and operator handles with an RAII guard so a failed encode
    does not leak
  - move the NPU runtime (loaded session + dispatch I/O) into a public
    executors::cann
  - dequantize float16 filter constants correctly and bounds-check the OIHW
    transpose
  - reject ScatterND (the DDK's ScatterNdUpdate aborts the process on-device) 
  - extend the test_cann_execution device test with an Add re-dispatch case
  - sync cann_expected_failures.txt from a full device run (2317 -> 1729 entries;
    792 passing trials)

## Validation

- [ ] `make test`

## Documentation


src/backends/cann.rs
```diff
@@ -199,74 +208,148 @@ impl<'context> MLBackendContext<'context> for CannContext {
             });
         };
 
-        let model_bytes = &cann_graph.model_bytes;
+        let t0 = std::time::Instant::now();
```
theHamsta, line 211: are those simple time stamps enough for you right now or would you like to have something more sophisticated like https://docs.rs/tracing/latest/tracing/ that can be connected to all kinds of profiler, logging, telemetry outputs?

src/converters/cann.rs
```diff
@@ -89,9 +90,10 @@ pub(crate) fn webnn_op_to_hiai(op: &Operation) -> Option<&'static str> {
 
         // ── Convolution + Pool + Matmul ──────────────────────────────
         Operation::Conv2d { .. } => Some("Conv2D"),
-        Operation::ConvTranspose2d { .. } => Some("ConvTranspose"),
+        Operation::ConvTranspose2d { .. } => Some("Conv2D"),
```
theHamsta, line 93: ConvTranspose2d is `Conv2D`?

src/converters/cann.rs
```diff
@@ -494,6 +1024,35 @@ mod adapter {
...
+            }
+        }
+
+        // Developer aid: `RUSTNN_DEBUG=2` dumps every op's shapes (on-device).
+        if std::env::var("RUSTNN_DEBUG")
+            .map(|v| v == "2")
+            .unwrap_or(false)
+        {
+            dump_op_shapes(graph);
+        }
```
theHamsta, line 1054: if you want you can put those under `trace!` and set a manual log category for those shapes

https://github.com/rustnn/trtx-rs/blob/598a6076cf789df3ac638ae14fcf8f0c7d68dc17/trtx/src/logger.rs#L45-L45

if you are using env_logger you can set a different logging level for each category https://docs.rs/env_logger/latest/env_logger/ or raise the logging verbosity of `rustnn::converters::cann`

src/backends/cann.rs
shubhamg13: Keeping the CANN_TIMING timestamps for now. tracing infrastructure is planned as a follow-up (starting in mlgraphbuilder), and these will move over then.

src/converters/cann.rs
shubhamg13: Yes, intentional: native ConvTranspose is unsupported on the NPU, so ConvTranspose2d is lowered to a decomposed Conv2D (stride=1 + recomputed pads). Added a one-line note on the mapping.

src/converters/cann.rs
shubhamg13: Deferred to the tracing follow-ups: these will move to log::trace! with the rustnn::converters::cann category instead of the RUSTNN_DEBUG env gate.

src/backends/cann.rs
shubhamg13: That's a nice idea. I would like to approach this separately. 

Firstly, wiring trace in common backend and then in specific backends, WDYS?

src/converters/cann.rs
shubhamg13: At first, I also got lost. 

Yeah, it needs additional comment to avoid further confusion.

src/backends/cann.rs
theHamsta: sounds good!

shubhamg13:
I will upload expectations soon, rest of logic is open for review.

Edit: Ready for Review

mtavenrath (dismissed):

PR #238, opened 2026-09-18
Unify shape inference and enable executable WebNN reloads

FelixKrall:
## Summary

RustNN changes to enable full roundtrip numeric validation of onnx2webnn conversion. Tied to [onnx2webnn #5](https://github.com/rustnn/onnx2webnn/pull/5) and dependend on [webnn-graph #19](https://github.com/rustnn/webnn-graph/pull/19). The reference in this PR needs to updated after the webnn-graph was merged.

Make serialized WebNN graphs independently reloadable and executable while unifying graph recording and shape inference between `MLGraphBuilder` and the `.webnn` loader.

Completed graphs now use an unambiguous shape model: `[]` always means a known rank-zero scalar, bounded dynamic dimensions remain explicit, and unresolved descriptors exist only as temporary internal inference state.

## Feature additions

- Add `MLContext::rustnn_build_graph` as a direct compilation entry point for deserialized `GraphInfo`. This avoids constructing a second, unused `GraphRecorder` when compiling a graph reconstructed by the `.webnn` loader.
- Add `run_onnx_path_with_inputs` so native ONNX Runtime execution can resolve external-data sidecars relative to the model file. This supports reference validation of filesystem-backed models whose weights are not embedded in the ONNX protobuf.
- Serialize complete operation arguments, including reshape targets, Slice parameters, concat axes, permutations, and operand-valued options referenced by stable names.
- Reload serialized graphs through the same recording and inference path used by `MLGraphBuilder`.
- Add a versioned packed-4-bit Safetensors extension:
  - Logical `Int4` and `Uint4` dtype and shape remain in `.webnn`.
  - Packed low-nibble-first bytes are stored as Safetensors `U8`.
  - Archive metadata identifies the RustNN extension.
  - Reload validates marker, dtype, shape, byte length, and tensor-name resolution.
- Support mixed ordinary and packed-4-bit constants in one Safetensors archive.

## Bugfixes

- Infer and record every output of multi-output operations rather than only the first.
- Prevent unresolved operands from silently becoming scalar descriptors.
- Treat scalar GRU hidden states as rank-zero values and reject them through normal GRU rank validation.
- Preserve scalar inputs, constants, intermediates, outputs, quantize/dequantize operands, and converter shape-map entries.
- Always serialize required scalar `Reshape` and `Expand` targets as `newShape: []`.
- Reject missing or malformed required shape-valued arguments explicitly.
- Serialize operand-valued options by stable name instead of unstable numeric operand IDs.
- Validate packed 4-bit logical element counts and exact packed storage lengths.
- Correct Slice lowering for non-unit strides by deriving backend end indices from `start + extent`; this fixes both the ONNX and LiteRT lowering paths.

## Refactors

- Introduce a private shared `GraphRecorder` used by both graph construction and JSON loading.
- Centralize descriptor inference, operation insertion, dependency tracking, and graph-output marking.
- Make operation insertion atomic: failed inference or validation no longer leaves partially recorded graph state.
- Remove the historical empty-vector unknown-shape heuristic and obsolete temporary shape-table workaround.
- Keep unresolved shapes represented through `Option` or missing internal entries rather than public `OperandDescriptor` values.
- Update setup documentation with concise optional prerequisites for feature-specific backends.

## Behavioral impact and compatibility

- `GraphRecorder` remains crate-private and is not a new public API.
- `MLContext::rustnn_build_graph` and `run_onnx_path_with_inputs` are additive public entry points.
- Every `OperandDescriptor` in completed `GraphInfo` now has a known shape:
  - `[]` is scalar.
  - Nonempty shapes may contain bounded `Dimension::Dynamic` entries.
  - Unknown shapes cannot escape graph construction.
- The experimental `.webnn` format has intentional compatibility breaks:
  - Required shapes may no longer be omitted.
  - Scalar shape arguments must be serialized explicitly.
  - Direct input or constant graph outputs are rejected; an explicit operation such as Identity is required.
- Ordinary Safetensors remain unchanged. Packed 4-bit tensors require the RustNN metadata marker and are rejected if presented as an unmarked or malformed extension.
- Browser WebNN still does not natively execute 4-bit tensors. The ORT backend reconstructs an executable graph using its supported representations.

## Validation

- Latest `cargo test --lib`: 379 tests passed.
- Default, no-default-feature, `dynamic-inputs`, ONNX Runtime, and available backend-mock checks passed across the refactor.
- Focused round trips passed for:
  - Scalar and bounded-dynamic descriptors.
  - Reshape, Expand, Slice, Concat, and Gemm bias.
  - Multi-output inference.
  - Even- and odd-sized `Int4` and `Uint4` constants.
  - Mixed native and packed-4-bit Safetensors.
  - Packed-4-bit graph compilation and ORT execution.
  - Malformed packed archive rejection.
- Downstream onnx2webnn validation passed 743 tests and completed the 52-case skeleton and real-weight sweeps with the expected documented blockers.
- Formatting and diff checks passed.
- LiteRT regression coverage was added, but the LiteRT feature test could not run locally because `flatc` was unavailable.
- Clippy remained blocked by pre-existing `-D warnings` failures outside this change.


src/graph_recorder.rs
```diff
@@ -0,0 +1,528 @@
...
+///
+/// Operation insertion is atomic: inference and output validation happen before
+/// the operation or any of its output operands are appended to `GraphInfo`.
+#[derive(Clone, Debug, Default)]
+pub(crate) struct GraphRecorder {
+    graph: GraphInfo,
+}
+
+impl GraphRecorder {
+    pub(crate) fn new() -> Self {
```
theHamsta, line 19: to me it seems as if all those should be methods on GraphInfo and some of them slowly replacing the public access to fields of GraphInfo

theHamsta: for me fine, to have this struct first. I can compare with the changes I made to GraphInfo. But long-term a manageable interface like this should replace direct data access GraphInfo

src/operator_options.rs
```diff
@@ -423,7 +418,7 @@ pub struct MLConstantOptions {
     pub data: Option<String>, // base64
     pub data_type: String,
     #[serde(default)]
-    pub shape: Vec<u32>,
+    pub shape: Option<Vec<u32>>,
```
theHamsta, line 421: what is the meaning of shape being `None`?

src/loader.rs
```diff
@@ -84,16 +88,185 @@ fn map_weight_resolve_error(err: WeightResolveError) -> GraphError {
...
+        format: "safetensors-packed4".to_string(),
+        reason: reason.into(),
+    }
+}
+
+fn sanitize_weight_key(name: &str) -> String {
+    name.replace("::", "__").replace('.', "_")
+}
+
+fn discover_external_weights(graph_path: &Path) -> Option<PathBuf> {
```
mklimenko-nv: This is a copy-paste from webnn-graph. Can we avoid it and come to a single source of truth?

src/loader.rs
```diff
@@ -84,16 +88,185 @@ fn map_weight_resolve_error(err: WeightResolveError) -> GraphError {
...
+    }
+
+    let Some(weights_path) = discover_external_weights(graph_path) else {
+        return Ok(());
+    };
+    if !is_safetensors_path(&weights_path) {
+        return Ok(());
+    }
+
+    let bytes = fs::read(&weights_path).map_err(|source| GraphError::io(&weights_path, source))?;
```
mklimenko-nv: Here and below: the whole archive is read (potentially several gigabytes), the header is parsed, then everything is copied into `InlineBytes` at line 252 and the buffer is dropped.

src/loader.rs
```diff
@@ -130,6 +303,7 @@ pub fn load_graph_from_path(path: impl AsRef<Path>) -> Result<GraphInfo, GraphEr
         });
     };
 
+    resolve_packed_4bit_weights(&mut graph_json, path_ref)?;
     webnn_graph::external_weights::resolve_external_weights(&mut graph_json, path_ref, None, None)
```
mklimenko-nv, line 132: `webnn_graph` resolver discovers the file again and reads the full buffer again for the remaining tensors.

mklimenko-nv (dismissed):
Aside from those comments, the upstream documentation has no page that describes the `.webnn` or weights format at all. Could you make a pass on the docs for your changes as well?

src/graph_recorder.rs
FelixKrall: Yes I agree, this is a first attempt to guide the project into this direction. Its motivation was rather to unify shape inference in .webnn loading and the Graph construction path that is used when using the builder API or the .webnn loader. 
This does not make GraphInfo crate private yet, but it creates an implementation entry point to adopt when doing this in the future.

src/operator_options.rs
FelixKrall: None means the required constant shape was omitted during parsing or construction; it is not an unknown tensor shape. We need the distinction because Some(vec![]) is a valid rank-0 scalar. constant_shape() rejects None before the operation enters GraphInfo, so every completed operand still has a known shape.

FelixKrall:
@mklimenko-nv thanks for catching the memory and code duplication issues. I was not aware of the webnn-graph ownership of this. I'll fix it ASAP and also include the docs you are requesting.

FelixKrall:
I have addressed your comments as such @mklimenko-nv.

- webnn-graph is now the single source of truth for sidecar discovery and @weights resolution.
- Added support for int4/uint4 for .safetensors in the .webnn contract (like this rustnn branch previously implemented). Because SafeTensors has no native 4-bit dtype, the logical dtype and shape remain in .webnn, while the packed bytes are stored as U8 with the versioned rustnn.webnn.packed4=1 metadata marker.
- This allowed the duplicate load to disappear form RustNN because all weight refs are now handled in a single pass by the same function. The external weight resolver is invoked here: ([RustNN call site](https://github.com/FelixKrall/rustnn/blob/06f7aefcc4d4e3a7c68bee75ec2aa88861698544/src/loader.rs#L94-L135))
- Changed the .safetensors loads to use mmap from disk instead of buffering reads like they previously did for speedup - see [mmap helper and SafeTensors use](https://github.com/FelixKrall/webnn-graph/blob/32a685c0cc006e8b1d5c3e2a8499fc173cb5997c/src/external_weights.rs#L217-L245)). If you don't like this, let me know.

I have also added docs on webnn-graph usage here and about its current conventions in the [webnn-graph PR](https://github.com/rustnn/webnn-graph/pull/19). I removed the stale agent like docs from the webnn-graph repo as well.

mtavenrath:
In the long run I want to see weight loading with rusts https://docs.rs/io-uring/latest/io_uring/ for real performance. mmap on Windows and DGX Spark tends to be extremely slow in comparison to (unbuffered) file io.

FelixKrall:
> In the long run I want to see weight loading with rusts https://docs.rs/io-uring/latest/io_uring/ for real performance. mmap on Windows and DGX Spark tends to be extremely slow in comparison to (unbuffered) file io.

Makes sense, I reverted that but back to the previously in place standard file i/o in webnn-graph.

FelixKrall:
Rebased onto rustnn main

src/operator_options.rs
theHamsta: but why is this in operator options? for WebNN all shapes are always known. This is public API of WebNN

FelixKrall: I agree this is incorrect, did not catch that. However MLConstantOptions is not public WebNN API itsself. It will later be rejected by constant_shape() in GraphRecoder as is and missing shapes in .webnn are not even legal anymore. This can not propagate to GraphInfo, it is effectively a dead code path.
It still should not be possible to create this as None, I can create a small cleanup PR for this.

Dou you think removing these non spec options alltogether is best?

theHamsta: I think you're right `MLConstantOptions` shouldn't be called that way. It does not exist in WebNN. I was just confused why one would ever have a constant with unknown shape.

I think this is an artifact from the onnx op `Constant` which can be used instead of initializers.

theHamsta: IHMO, we should get rid of the `"constant"` op. It is never used within WebNN itself and some converters even don't support it. It just came from onnx converters.

FelixKrall: Yes agreed, I think this being on main now is fine, but I will create a minimal follow up PR that tries to get rid of these OPs and at least the option to have them be initialized with None shape. They should't be and trying to should cause the error not the check later.

PR #248, opened 2026-09-23
fix(trtx): use select for triangular instead of multiply for NaN handling

theHamsta:
## Summary

This fixes the recently added WPT tests for `triangular` (4 failing tests)

## Validation

- [x] `make test`

## Documentation

See `docs/development/documentation-policy.md` for which page belongs to which change.

- [ ] Updated the rustdoc comments and docs pages that describe the changed behavior (or no documented behavior changed)
- [ ] `make docs-api` and `make ci-docs` pass if rustdoc or docs pages changed


src/converters/trtx.rs
```diff
@@ -14912,7 +14885,7 @@ impl TrtxConverter {
             .map(|s| get_static_or_max_size(s) as i64)
             .collect();
         let mask_layer = network
-            .add_constant_owned(&dims, mask_bytes, mask_trt_ty, None)
+            .add_constant_owned(&dims, mask_data, TrtDataType::kBOOL, None)
```
theHamsta: of course, even if a boolean buffer is better, it will still cause an overhead for very big matrices (e.g. causal attention). Currently, I don't know of a way to express the triangular matrix implicitly. We might need to pattern match against attention patterns.

